PR #1, opened 2020-01-31
added a missing __init__.py file, blocking import

marius851000:
other wise it fail to import


Parakoopa:
Thanks! My local Python doesn't really mind if the __init__ files are missing, but I guess it depends on the setup.

PR #24, opened 2020-11-11
Support for WTU/WTE files

irdkwia:
I have improved support for WTU and WTE files.
- Now, it allows different image modes (Palette-only, 2bpp, 4bpp and 8bpp) and actual dimensions when importing and drawing.
- It provides support for importing and drawing WTE images without palette too.
- I also updated the WTU file to explicit all the unknown values.

Merging with fixed-floors because skytemple's master branch currently imports elements that are only defined in this branch.
This pull request and Skytemple/skytemple#57 should fix Skytemple/skytemple#56.


Parakoopa:
Thank you!! Don't be confused about the commits, I just rebased on master.

PR #25, opened 2020-11-12
Improving fixed floors and monster spawns types.

irdkwia:
I have changed some things about fixed floors:
- I created an enumeration for enemy settings/monster spawn types, because there seems to be more types than just enemies/allies.
- The unknown boolean 10 in fixed floor properties determines if you automatically exit the floor when you defeat all the enemies on the floor (like in a boss fight).
- Some other minor fixes.

Pull request Skytemple/skytemple#58 must also be accepted to work.


irdkwia:
Just fixing a mistake about the monster types enum: some values were used by the game but not defined in my enum, so there was an error on that.

PR #57, opened 2021-01-02
Add support for chr files + cartridge removed image

irdkwia:
Adds model/handler/writer for editing chr files (used for upper screen window styles).
Adds module to edit the hardcoded cartridge removed image in arm9.


irdkwia:
Also, I'm fixing one issue I noticed with fixed rooms, when resizing it inserted a blank row of rules at the top.

PR #58, opened 2021-01-04
wan read/write for overworld sprites

audinowho:
Draft pull request open for comment.

There are 6 major functions in this code, all currently in writer.py:

ImportWan - takes a file-like, returns a WanFile.  Should it take the whole byte array instead?
ExportWan - takes a WanFile and file-like, writes to the file-like
ImportSheets - takes a directory path, returns a WanFile.  Currently lacks a ton of validation steps, and throws an exception if it reads more than 16 colors.  Should it take a list of PIL images and an xml doc instead?
ExportSheets - takes a wan, writes spritesheets to the destination.
MergeWan - Takes multiple wan files (ex, the wans from monster, m_attack, and m_ground of the same pokemon) and returns a single wan with animations from all put together.  Assumes all 3 wan files have the same animation for each index and the same palette (this is true for all sprites in EoS).
SplitWan - Takes a single wan file and divvies up the animations between several wan files based on the arguments.  Used to write back wans from a single sheet.

Not sure if I should just move the import/export methods into model.py, or if the same should be done for other helper functions...


skytemple_files/graphics/chara_wan/dbg/test.py
```diff
@@ -0,0 +1,63 @@
+import sys
+import os
+import shutil
+import math
+import glob
+import chara_wan.writer as exWriter
+
+
+
+def getWanFilePath(baseDir, dirName, index):
```
Parakoopa, line 10: Can you please use `snake_case` for all functions, variables and methods? If you want I can probably auto-convert it.

Parakoopa:
First again thank you so much for this!

skytemple_files/graphics/chara_wan/dbg/test.py
```diff
@@ -0,0 +1,63 @@
...
+    for file in glob.glob(os.path.join(inDir, '*_' + format(index, '04d') + '_*')):
+        inFile = file
+    return inFile
+
+def ConvertJoinSeparateChara(baseDir, index, anim_name_map, sdwImg):
+    monsterPath = getWanFilePath(baseDir, 'monster', index)
+    m_groundPath = getWanFilePath(baseDir, 'm_ground', index)
+    m_attackPath = getWanFilePath(baseDir, 'm_attack', index)
+
+    print(format(index, '04d'))
```
Parakoopa, line 22: Please remove prints and replace them with a logger instead.
To add a logger, just add the following to the very top of each module you need one:
```py
import logging
logger = logging.getLogger(__name__)
```

skytemple_files/graphics/chara_wan/dbg/test.py
```diff
@@ -0,0 +1,63 @@
+import sys
```
Parakoopa, line 1: Please add a copyright header to all files, see the other files in the repo.

skytemple_files/graphics/chara_wan/model.py
```diff
@@ -0,0 +1,310 @@
...
+
+ATTR01_ResMask = 0xC000  # 1100 0000 0000 0000
+
+TEX_SIZE = 8
+DIM_TABLE = [(1, 1), (2, 2), (4, 4), (8, 8), \
+             (2, 1), (4, 1), (4, 2), (8, 4), \
+             (1, 2), (1, 4), (2, 4), (4, 8)]
+
+
+class WanFile(object):
```
Parakoopa, line 47: `(object)` is deprecated.

skytemple_files/graphics/chara_wan/writer.py
```diff
@@ -0,0 +1,1412 @@
...
+              21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43]
+
+DEBUG_PRINT = False
+
+
+
+
+
+
+def ExportWan(out_file, wan):
```
Parakoopa, line 29: Please make this return a byte array instead / work on a bytarray. You can use the `io` module if you want to use pseudo files.
Otherwise the `util` module has utility functions for working with byte arrays:
https://github.com/SkyTemple/skytemple-files/blob/master/skytemple_files/common/util.py

skytemple_files/graphics/chara_wan/writer.py
```diff
@@ -0,0 +1,1412 @@
...
+    out_file.write(ptrWAN.to_bytes(4, 'little'))
+    out_file.write(ptrSir0.to_bytes(4, 'little'))
+
+
+# This will accurately load all sir0 found in m_ground, m_attack, and monster.bin with a few exceptions:
+# m_ground_0546_0xb01840.wan - Armaldo.  Metaframe Unk#0 is a nonzero
+# m_ground_0587_0xbd0a10.wan - Latios.  Unknown difference
+# monster_0433_0x3dd1c0.sir0 - Giratina Origin.  Extra zeroes after file end.
+# monster_0438_0x3ec6d0.sir0 - Shaymin Sky.  Extra zeroes after file end.
+def ImportWan(in_file):
```
Parakoopa, line 267: same as above, please accept bytes as input instead.
Also please use the existing Sir0 logic and if you want the `Sir0Serializable` base class for the model to make that more straight forward. Same applies to import.
It would probably also be better to move the read-in code to another module.

skytemple_files/graphics/chara_wan/writer.py
```diff
@@ -0,0 +1,1412 @@
...
+                else:
+                    animSequence.append(SequenceFrame(frameIndex, frameDur, flag,
+                                                    (sprOffX, sprOffY), (sdwOffX, sdwOffY)))
+            animGroup.append(animSequence)
+        wan.animGroupData.append(animGroup)
+
+    return wan
+
+
+def ImportSheets(inDir):
```
Parakoopa, line 503: Please move that to a separate module. For portaits I named it `sprite_bot_sheet.py` but you can also come up with a better name :P

skytemple_files/graphics/chara_wan/writer.py
```diff
@@ -0,0 +1,1412 @@
...
+            shadow_img = Image.open(os.path.join(inDir, anim_name + '-Shadow.png')).convert("RGBA")
+
+            anim_sheets[index] = (anim_img, offset_img, shadow_img)
+
+    # raise warning if there exist anim stats without anims, or anims without anim stats
+    if len(anim_names) > 0:
+        orphans = []
+        for k in anim_names:
+            orphans.append(k)
+        raise Exception("Xml found with no sheet: {0}".format(', '.join(orphans)))
```
Parakoopa, line 585: In general please try to use more specific exceptions, such as `ValueError`, `KeyError` or `RuntimeError`, etc.

skytemple_files/graphics/chara_wan/writer.py
```diff
@@ -0,0 +1,1412 @@
...
+        for copy_idx in copies:
+            animGroupData[copy_idx] = duplicateAnimGroup(animGroupData[idx])
+
+    wan = WanFile(imgData, frameData, animGroupData, offsetData, [singlePalette])
+    wan.sdwSize = sdwSize
+    # return the wan file
+    return wan
+
+
+def ExportSheets(outDir, sdwImg, effectData, anim_name_map):
```
Parakoopa, line 783: same, just add that to the sheet module as well.

Parakoopa (commented):
Please add a handler to the module, see https://github.com/SkyTemple/skytemple-files/blob/master/skytemple_files/graphics/kao/handler.py.

Parakoopa (commented):
See review comments.

skytemple_files/graphics/chara_wan/dbg/test.py
Parakoopa: Nevermind, print is okay in the dbg script.

Parakoopa:
Looks fine to me now except for the casing (but I can go over that automatically when the rest is done, so no big issue here) and the Sir0 thing we talked about. I'll also try and add it to the UI later today or tomorrow as a test.

PR #78, opened 2021-02-16
Cipnit's asm hacks

MkTr:
Contains the .asm files for the following hacks:
-Complete Team Control v1.2
-Partners Trigger Hidden Traps v1
-Far-Off Pal Overdrive v0
-Reduce Jumpcut Pause Time v1
Also comes with descriptions to put in the python files for the hacks.


skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+.pool
+
+@DequipPenalty:
+	ldr r0,=@CurrentLeader
+	ldr r0,[r0]
+	cmp r0,r4
+	mov r0,0h
+	moveq r0,1h
+	bx r14
+.pool
```
End45, line 136: You could use a single function since they are the same

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+@WaitChecker:		;This skips a pokemon's turn if the pokemon swapped places with another pokemon previously. In the original game, swapping places with a partner skips the partner's next turn, as if the partner "used its move" when it swapped places with you. (A bug in the game raises this skip flag for the leader too, which I nopped earlier.) The game has nothing to prevent a leader from moving if this flag is raised, so this exists to make the game play as it's supposed to.
+	push r1-r3,r14
+	ldrsh r0,[r0,r1]
+	cmp r0,0h
+	popeq r1-r3,r15		;If this is equal to 0, the Pokemon wasn't going to take a turn anyway
+	ldr r1,[r4,0B4h]	;r4 has the current pokemon's entity address
+	ldrh r2,[r1]
+	mov r3,4000h
+	and r3,r2,r3
+	cmp r3,0h
```
End45, line 204:
```suggestion
	tst r2,4000h
```

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+	push r14
+	ldr r0,[r7,0B4h]
+	mov r1,0h
+	strb r1,[r0,7h]
+	ldr r0,[r6,0B4h]
+	mov r1,1h
+	strb r1,[r0,7h]
+	str r6,[r5]
+	pop r0
+	bx r0
```
End45, line 220:
```suggestion
	pop r15
```

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+	mov r8,0h
+	ldr r12,=NA_02301D10
+	bl NA_022EBD10		;jumping to the start of the round function
+roundover:
+	mov r6,r8
+	bl @leaderchanger	;first pokemon gets to be the leader again
+	pop r0-r12
+	add r13,r13,4h
+	mov r3,0h
+	bl NA_022EC094		;skipping the turns of partners
```
End45, line 295:
```suggestion
	b NA_022EC094		;skipping the turns of partners
```
Just to make clear that we aren't going to return here

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+.close
+
+.open "overlay_0031.bin", ov_31
+.org NA_02387530	;Jump to team submenu function
+	bl @paranoia_agent	;original: bl NA_022EB408
+.close
+
+
+.open "overlay_0036.bin", ov_36
+.org CompleteTeamControl_RAMPosition
```
End45, line 60:
```suggestion
.orga 0x10
.area 3F0h - 10h
```
No need to know the start offset of the extra overlay.
I'll leave 16 reserved bytes at the start of the file so we can add a header if it's ever needed, you can have 10h - 3F0h.
If you want some extra bytes just in case you need to make changes later let me know. The allocated space for your hack can be changed until somebody else requests another area inside the overlay.

End45 (changes requested):
Make sure to test this before it's merged! I suggest testing with a clean ROM and then with a ROM that has Psy's ActorAndLevelListLoader patch applied, so you can be sure that there aren't any hardcoded RAM addresses left.
I added some improvements. The size of the extra code area will decrease if you apply them, but we can keep the original amount if you want so you have some free space in case you need to add some more instructions to fix something in the future.

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
MkTr: Removing the `mov` operation is a problem - r3 is later used to subtract r2, so this leads to undefined behavior. I'll replace `and` and `cmp` though.

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+	cmp r2,0h
+	mov r2,0h		;changing to auto
+	moveq r2,1h		;changing to manual
+	strb r2,[r1]
+	ldr r1,=@Message_SetToAuto
+	addeq r1,r1,@Message_SetToManual -@Message_SetToAuto
+	mov r0,r4
+	mov r2,1h
+	mov r3,0h
+	mov r5,r12	;i don't know if it's ok to do this
```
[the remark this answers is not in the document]
MkTr: I, uh... I forgot why I wrote that. I think I was suspicious r5 was used as a parameter for the message function, or I thought r12 had to be set to something specific before the message function. idk I'll just push r12

skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
```diff
@@ -0,0 +1,299 @@
...
+	mov r2,20h
+	strb r2,[r1]
+	sub r1,r1,1h	;=@ManualModeOn
+	ldrb r2,[r1]
+	cmp r2,0h
+	mov r2,0h		;changing to auto
+	moveq r2,1h		;changing to manual
+	strb r2,[r1]
+	ldr r1,=@Message_SetToAuto
+	addeq r1,r1,@Message_SetToManual -@Message_SetToAuto
```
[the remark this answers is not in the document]
MkTr: This actually adds space lol - `ldr rX,=Y` operations take 8 bytes, 4 for the operation itself and 4 for the value which is put in the pool data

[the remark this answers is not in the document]
End45: Eh, fair. I added this so it's a little bit clearer, but you can keep it this way too.

End45 (commented):
Looks good. The only thing left would be the Python files. You can see how Para did it with my patches some time ago here: https://github.com/SkyTemple/skytemple-files/commit/da35ca6bc9290400cdf11191436e392920b0576a
Part of the process includes reading a value that is changed by the patch somewhere to see if it has been applied already. You can use any value or instruction from the original game that changes after it's applied.

Parakoopa:
The definitions in the `pmdscriptdata.xml` are also missing, but they are also in that commit. Since your patches select the region the same way as for End's patches, you can just use that basically.

MkTr:
I'm not familiar with python, but yeah, I'll give it a shot.

MkTr:
Alright - I don't know if CompleteTeamControl.py works, but I think everything else is right. I don't know if I missed anything, I updated as much as I could find.

PR #79, opened 2021-02-18
Adds a patch to change the implementation of the move effect table.

irdkwia:
Adds a patch to delete the move effect switch in overlay 29 and instead load the effect ASM code directly from the ROM file system and jump to it. 
WARNING! This is completely experimental!
This assumes that there is no recursive call of the function that executes moves. Since this wasn't tested with all the moves, there may be a move I haven't thought of that needs to execute another.
Luckily, I tested some of the ones that could have had a recursive call (Mimic based moves, Me First and Metronome) and it seems that it doesn't have any problem with it. 
It may be unstable on real hardware. It seems to work when manipulating the cache "properly" though.
Also adds support for editing the new file in the ROM file system.


Parakoopa:
Force-push was a rebase.

PR #86, opened 2021-03-02
Add method to import/export sheets with zip

riiswa:
Reference to https://github.com/SkyTemple/skytemple/issues/167

This new method allows saving the sheets in a zip archive. I was able to do it on my local version of skytemple. The next step will be to add the button to the application.

Here is the generated archive:
[bulbasaur.zip](https://github.com/SkyTemple/skytemple-files/files/6065362/bulbasaur.zip)


riiswa:
I will try to do the same thing with the import

riiswa:
If it's better, I can reuse the logic of creating the temporary directory for export.

Parakoopa:
Thank you! Yes that would be better!

PR #92, opened 2021-03-12
Add patches to implement Fairy type and Fairy gummies.

irdkwia:
Add patches to implement the Fairy type: 
- A patch to change the type match-up table implementation and add Fairy type as the "Neutral" (this replaces the old table with the 6th+ gens table) and all tables related to Gummi effects by type (IQ boosts/Belly recover).
- A patch to transform unused item 138 into a Fairy-type Gummi.

To create those patches, some new files have been created: 
- Handler/Model/Writer for file "BALANCE/item_p.bin".
- A patch to extract Spinda bar's item list.
- Handler/Model/Writer for file "BALANCE/itembar.bin" (the file created by the patch mentioned in the previous point).

### DIFF
--- a/skytemple_files/dungeon_data/mappa_bin/item_list.py
+++ b/skytemple_files/dungeon_data/mappa_bin/item_list.py
@@ -41,7 +41,7 @@ class MappaItemCategory(Enum):
     THROWN_PIERCE = 0, 1, 6, [], [9]
     THROWN_ROCK = 1, 7, 4, [9], []
     BERRIES_SEEDS_VITAMINS = 2, 69, 40, [], [116, 117, 118]
-    FOODS_GUMMIES = 3, 109, 29, [116, 117, 118], []
+    FOODS_GUMMIES = 3, 109, 30, [116, 117, 118], []
     HOLD = 4, 13, 56, [], []
     TMS = 5, 188, 105, [], []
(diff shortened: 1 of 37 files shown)


skytemple_files/dungeon_data/mappa_bin/item_list.py
```diff
@@ -41,7 +41,7 @@ class MappaItemCategory(Enum):
     THROWN_PIERCE = 0, 1, 6, [], [9]
     THROWN_ROCK = 1, 7, 4, [9], []
     BERRIES_SEEDS_VITAMINS = 2, 69, 40, [], [116, 117, 118]
-    FOODS_GUMMIES = 3, 109, 29, [116, 117, 118], []
+    FOODS_GUMMIES = 3, 109, 30, [116, 117, 118], []
```
End45, line 44: Could this cause issues when the patch isn't applied?

Parakoopa: Probably not, since item # 138 isn't in any other category. It may cause slight issues when adding it to a floor before applying the patch, but the impact should be minimal. Could maybe add check to the UI, but idk if it's worth it.

Parakoopa (approved):

PR #96, opened 2021-03-24
Adds a patch for an extra menu to choose the starter

irdkwia:
Adds a patch which lets the player discard the choice made by the personality quiz and choose the starter instead.
Seems to work on real hardware.
Extends overlay 13 size by 0x1000 bytes (should not be a problem since the size of this overlay still is below the size of some other menu overlays (e.g. 19)).

### DIFF
--- a/skytemple_files/patch/handler/choose_starter.py
+++ b/skytemple_files/patch/handler/choose_starter.py
@@ -1,0 +1,116 @@
+#  Copyright 2020-2021 Parakoopa and the SkyTemple Contributors
+#
+#  This file is part of SkyTemple.
+#
+#  SkyTemple is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  SkyTemple is distributed in the hope that it will be useful,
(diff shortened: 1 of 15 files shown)


irdkwia:
Adds a patch to be patched over the previous one to skip the quiz part and directly go to the menu to choose the partner.
The behavior for the ribbon / items you get is not tested yet (should be either default or the one that corresponds to the selected pokémon's personality).

skytemple_files/patch/handler/choose_starter.py
```diff
@@ -40,13 +40,9 @@
...
 
-# TODO; support other languages
-DIAG_LIST = {"MESSAGE/text_e.str": "Then, who would you like to be?",
-             "MESSAGE/text_f.str": "Quel pokémon souhaitez-vous devenir?",
-             "MESSAGE/text_g.str": "---",
-             "MESSAGE/text_i.str": "---",
-             "MESSAGE/text_s.str": "---",
-             "MESSAGE/text_j.str": "---"}
+MESSAGE = "Then, who would you like to be?"
+# For xgettext scanning: _("Quel pokémon souhaitez-vous devenir?")
```
Parakoopa: fixed in next commit

PR #108, opened 2021-04-24
Extend the Number of Pokémon entries

irdkwia:
Pull requests #104, #105, #106 need to be merged before that
EXPERIMENTAL! Always make a backup before there is a 99.99% chance I forgot something that will break the game and changes can't be reverted
Tests are needed for almost everything in the game

If/When this works properly, extends the number of possible Pokémon slots from 600 entries with 2 possible forms to 2048 independent entries
Tries to maintain consistency upon applying the patch, this includes: 

[ASM]: 
- all the data access function for movesets, monster properties and portraits have been changed to work with the new system
- special monster personalities have been moved after the new entries
- special fixed floor monster IDs have been moved after the new entries IDs
- the code that spawned 2nd forms in even floors has been removed
- missions and Spinda bar can theoretically spawn new entries: unused flags in monster.md were turned into flags that sets mission attributes for each Pokémon
- recruitment search menu looks for the new entries
- the Adventure Log functions looks for the new Pokédex IDs up to 1184

[Files]: 
- waza_p.bin/waza_p2.bin movesets are extended and slots that where originally 2nd forms have a copy of the 1st form movesets
- same for m_level.bin level curves
- tbl_talk.tlk is also extended to contain new entries personalities
- kaomado is extended and missing portraits for 2nd forms are filled with 1st forms when available
- Pokémon Names and Categories have been moved to the end of the *.str files
- the newly md_evo.bin created file from patch #104 is also extended for new entries evolutions
- special Pokémon indices (>=1155) for fixed floor properties have been moved
- mappa_s.bin is reworked to explicitly contain female entries if they are valid on even floors, like they used to
- obviously, monster.md is extended to contain more entries; also, the hardcoded arm9 table MonsterSpriteData that contained dungeon sprite size and sprite file size values have been moved to the unk17 unk18 properties of that file (which are unused/don't have any code to access them); property entid now contains the id of the 1st form of an entry, used for many checks in the game (was unused)

### DIFF
--- a/skytemple_files/_resources/patches/asm_patches/irdkwia_asm_mods/expand_poke_list/common/fs_mappa_edited.asm
+++ b/skytemple_files/_resources/patches/asm_patches/irdkwia_asm_mods/expand_poke_list/common/fs_mappa_edited.asm
@@ -1,0 +1,650 @@
+; For use with ARMIPS
+; 2021/04/14
+; For Explorers of Sky All Versions
+; ------------------------------------------------------------------------------
+; Use filestreams to partially load mappa files instead of loading entirely
+; ------------------------------------------------------------------------------
+
+;0x3C - 0x44 = Moves
+;0x50 - 0x98 = Filestream
+;0x98 - 0xAC = Header
+;0xAC - 0xC0 = Lists
+
+.org LoadMappaFileAttributes
+.area 0xA74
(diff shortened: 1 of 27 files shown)


Parakoopa:
Wow, that's amazing! Can't wait to test this!

irdkwia:
Before merging this and PR in skytemple, there is one thing I need to change to fix missions generation (did it but I need to do some polishing).

irdkwia:
I think it's good now.

PR #119, opened 2021-05-25
Allow editing icon banner

tech-ticks:


Parakoopa:
I'll probably add editing the icon as well before merging (also to the UI :) )

tech-ticks:
> I'll probably add editing the icon as well before merging (also to the UI :) )

Thanks! I don't have any experience with NDS graphics so this would probably have taken me a while.

Parakoopa (approved):

PR #124, opened 2021-06-07
Extract Animation Table Patch

irdkwia:
Adds a patch that extracts the hardcoded animation tables and put them in one file, BALANCE/anim.bin, for more flexibility.

Testing required for this, as usual.

BALANCE/anim.bin file specs: 
- 0x14 bytes header (5 pointers * 4 bytes, in that order: TrapTablePtr, ItemTablePtr, MoveTablePtr, GeneralTablePtr, SpecialMoveTablePtr; each table ends where the next table starts, except the last one which ends at EoF, tables may be 4 bytes aligned)
- TrapTable: 0x2 bytes / trap: animation_id of an animation in the GeneralTable for each trap
- ItemTable: 0x4 bytes / item: two animation_id of animations in the GeneralTable for each item; I don't remember which one is for which case
- MoveTable: same as the Move Animation Table described in the animation doc (Table 1)
- GeneralTable: same as the Global Animation Table described in the animation doc (Table 2)
- SpecialMoveTable: same as the Special Move Animation Table described in the animation doc (Table 3)


Parakoopa:
Thank you!!

Parakoopa:
Relates to skytemple/skytemple#274

PR #128, opened 2021-06-12
enhancment in the arm patcher for packaging

marius851000:
Does two things :
- fix the the error display. Before this fix, displaying the error in case of failure failed, as ``f`` is a file rather than the function
- Ensure the permission are correctly set after copying the file (the folder where Nix store it's data is marked as read-only).


Parakoopa:
Thanks!

marius851000:
2min 20 to include it... That was.

PR #132, opened 2021-06-14
Add AllowUnrecruitableMons patch

End45:
Adds a patch that allows recruiting pokémon that are hardcoded to be unrecruitable (The other 3 Deoxys forms and the regis).


Parakoopa:
Thanks!

PR #133, opened 2021-06-19
Fixed a bug in fs_mappa patch

irdkwia:
Fixed an invalid pointer access with guaranteed items.
Also modifies ExpandPokeList as it uses that.


Parakoopa:
Thanks!

PR #134, opened 2021-06-24
Update MoveShortcuts to display the move list when holding L

End45:
The new version of MoveShortcuts displays the move list while holding L. It also prevents you from choosing linked or invalid moves.
Credits to @irdkwia!

MoveShortcuts now depends on the ExtraSpace patch.


Parakoopa:
Thanks to both of you!

PR #135, opened 2021-07-01
Just saying it's a dependant patch

irdkwia:
Otherwise depends_on will never work as intended.


Parakoopa:
@End45 mystery solved! Haha

Thanks!

Issue #28, opened 2020-11-12
Harcoded: Dungeon guest Pokémon


Parakoopa:
See dungeon document

---
Beep, boop, I'm [a bot](https://github.com/Parakoopa/IssueBot)! This issue was created by `@Parakoopa` in `#skytemple-tester`.

Parakoopa:
@End45 Do you think we could find a good way to also make the dungeon IDs assignable to the entries in SkyTemple?

End45:
I've been looking into this, it might be possible to make a patch that replaces all the hardcoded ID checks with a table that maps dungeon IDs to additional pokémon entries which could then be edited with SkyTemple. This would also allow adding or removing entries, since removing the original code frees some space.
It would require some time to implement it though. Once my holidays start I could look into it.

End45:
Done in https://github.com/SkyTemple/skytemple/pull/310
@Parakoopa This should be closed now

PR #142, opened 2021-07-12
Add EditExtraPokemon patch

End45:
Adds a patch that un-hardcodes the way the game adds additional pokémon to the team when entering uncleared dungeons. This data will now be part of a table inside arm9.bin, making it possible to edit it with a hex editor.

### DIFF
--- a/skytemple_files/_resources/ppmdu_config/pmd2data.xml
+++ b/skytemple_files/_resources/ppmdu_config/pmd2data.xml
@@ -1208,6 +1208,14 @@
           </Replace>
         </SimplePatch>
 
+        <SimplePatch id="EditExtraPokemon">
+          <Include filename="end_asm_mods/src/EditExtraPokemon.asm"/>
+          <Replace filename="end_asm_mods/src/common/regionSelect.asm" regexp='_REGION equ "(\w+)"'>
+            <Game id="EoS_NA" replace='_REGION equ "US"'/>
+            <Game id="EoS_EU" replace='_REGION equ "EU"'/>
+          </Replace>
+        </SimplePatch>
+
       </Game>
     </Patches>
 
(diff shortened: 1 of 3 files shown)


Parakoopa:
Wait, so what does this patch do exactly? Does this relate to #28?

End45:
Yes, this would be the first step to allow editing this data with SkyTemple.
If you remember, the way the game adds guest pokémon to the team is hardcoded by dungeon ID, after this patch is applied it will now be based on the data on a table inside arm9.bin. So it will be possible to edit that table to change where those additional pokémon are added to the team, as well as which special restrictions certain dungeons have (remember the hardcoded Hidden Land restrictions? Editing the table added by this patch allows you to set which dungeons have them now).

The next thing to do would be adding support to edit this table in SkyTemple, as well as the list of guest pokémon, which has space for 12 more entries after the patch is applied. I'm going to try and see if I can at least do the non-UI part of that.

Parakoopa:
Oh, so this replaces that other tables and adds other fields, gotcha!
That's great!

Parakoopa:
Oh, so this replaces that other tables and adds other fields, gotcha!
That's great!

Parakoopa:
Is it the same address as the old one? I couldn't quite figure it out from the patch.

End45:
The original table containing additional pokémon data it's still at the same location.
The patch adds a new one containing extra dungeon data, which includes data about which guest pokémon should be added to the team in each dungeon and those extra restrictions I mentioned (I'll have to see what's the start address for this table, since it depends on where the instructions section of the patch ends).
And since after that there's some empty space, I thought that it could be used as an extension to the guest pokémon table so more entries can be added (like I said, there's space for 12 more there).

Parakoopa:
Alright, now I understand! Thanks!

PR #143, opened 2021-07-13
Fix BGP model: tile 0 must be blank

irdkwia:
The game uses tile 0 to clear backgrounds.

Caused that background glitch in EoD.


Parakoopa:
Whoops?

PR #145, opened 2021-07-14
Guest pokémon support

End45:
Related to #28. Adds support to read/write guest pokémon data (requires the EditExtraPokemon patch to be applied).
Feel free to point out mistakes or make suggestions, this is the first python script I've written that has more than 10 lines.

### DIFF
--- a/skytemple_files/hardcoded/guest_pokemon.py
+++ b/skytemple_files/hardcoded/guest_pokemon.py
@@ -125,6 +125,7 @@ class GuestPokemon(AutoString):
                    self.level == other.level and \
                    self.iq == other.iq and \
                    self.atk == other.atk and \
+                   self.sp_atk == other.sp_atk and \
                    self.def_ == other.def_ and \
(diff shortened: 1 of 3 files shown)


skytemple_files/hardcoded/guest_pokemon.py
```diff
@@ -0,0 +1,269 @@
...
+            return False
+        else:
+            return self.unk1 == other.unk1 and \
+                   self.poke_id == other.poke_id and \
+                   self.unk2 == other.unk2 and \
+                   self.moves == other.moves and \
+                   self.hp == other.hp and \
+                   self.level == other.level and \
+                   self.iq == other.iq and \
+                   self.atk == other.atk and \
```
irdkwia, line 127: Special Attack missing (also in is_null)

End45: Fixed, thanks!

PR #147, opened 2021-07-16
Support patch parameters

irdkwia:
- AddExperienceShare has a parameter to set the % of exp. shared
- ChangeTextBoxBackground has 4 parameters to directly set the textbox background color
- ChangeMoveStatsDisplay has an option to add the power/accuracy value next to the bar. It also has a parameter to set the start ID of used text icons in markfont.dat, which allows to move the new icons anywhere in the file. Note that it does not check if selected entries are empty, so make sure that the selected entries can be used before applying the patch.

### DIFF
--- a/skytemple_files/patch/handler/move_growth.py
+++ b/skytemple_files/patch/handler/move_growth.py
@@ -22,7 +22,7 @@ from ndspy.rom import NintendoDSRom
 from skytemple_files.common.util import *
 from skytemple_files.common.ppmdu_config.data import Pmd2Data, GAME_VERSION_EOS, GAME_REGION_US, GAME_REGION_EU
 from skytemple_files.patch.category import PatchCategory
-from skytemple_files.patch.handler.abstract import AbstractPatchHandler
+from skytemple_files.patch.handler.abstract import AbstractPatchHandler, DependantPatch
 from skytemple_files.graphics.fonts.graphic_font.handler import GraphicFontHandler
 from skytemple_files.data.waza_p.handler import WazaPHandler
 from skytemple_files.data.waza_p.model import WazaMoveCategory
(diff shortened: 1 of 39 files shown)


skytemple_files/patch/handler/move_growth.py
```diff
@@ -67,15 +64,15 @@
                5000,1,1,1,
                   0,5,5,1]
 
-class MoveGrowthPatchHandler(AbstractPatchHandler):
+class MoveGrowthPatchHandler(AbstractPatchHandler, DependantPatch):
```
Parakoopa, line 67: i think this is a mistake?

skytemple_files/patch/patches.py
```diff
@@ -92,7 +92,7 @@ class PatchType(Enum):
     EXTRACT_MOVE_CODE = ExtractMoveCodePatchHandler
     EXTRACT_ITEM_CODE = ExtractItemCodePatchHandler
     EXTRACT_SP_CODE = ExtractSPCodePatchHandler
-    MOVE_GROWTH = MoveGrowthPatchHandler
+    #MOVE_GROWTH = MoveGrowthPatchHandler
```
Parakoopa, line 95: and this probably too?

Parakoopa (changes requested):

skytemple_files/patch/handler/move_growth.py
irdkwia: What's the problem?

Parakoopa: Oh nevermind, I didn't see the `depends_on` that was already there in the review screen.

skytemple_files/patch/patches.py
irdkwia: I removed it from the list because it depends on a patch that won't be merged until 1.4.

Parakoopa:
I merged the other PR, can you re-enable the growth patch?

Parakoopa (approved):

PR #148, opened 2021-07-20
Implement dungeon interruptions

irdkwia:
Adds the patch to implement dungeon interruptions.
Interruptions are all defined in "BALANCE/inter_d.bin".
Currently only model class is done in skytemple-files.


Parakoopa:
You are awesome.

Parakoopa:
I will merge this for 1.4 once the UI is done or if you do the UI before that then I'll merge it for 1.3 :)

PR #154, opened 2021-07-29
Fix Special Episode PCs entries

irdkwia:
20=0x20?


Parakoopa:
how did i even manage to do this??? thanks!

PR #156, opened 2021-08-03
Add FixMemorySoftlock patch

End45:
Adds a patch that can help prevent the crashes that happen when the game runs out of memory when displaying sprites in cutscenes (this has been observed in roms that had custom sprites or had used the randomizer). Other glitches could happen instead, but it will probably stop the softlock.
The patch has no effect in the normal behaviour of the game, as long as it doesn't run out of memory the patch will not trigger.
In the only test I've done so far, the game went from a softlock to running with no apparent side effects after applying the patch.


Parakoopa:
Thanks!

PR #157, opened 2021-08-07
Add ChangeFixedFloorProperties patch

irdkwia:
This patch removes hard-coded checks for fixed floor properties and replaces some with new properties.


Parakoopa:
Once again, thanks!

PR #167, opened 2021-09-07
Support for IQ groups' skills

### LINKED ISSUE #165: Edit IQ skills for IQ groups
@End45 will create a new patch for be able to assign more than 25 IQ skills per IQ group and make editing easier. After that is done, implement it into Files and the UI.

TBD for SkyTemple 1.3.1

End45:
This code can be used to read/write the list of IQ skills in each group. Reads support the uncompressed (vanilla) format and the compressed (the one used by my CompressIQData patch) format. Writes only support the compressed format since writing data in the original format won't be necessary.

I still have to add the patch to SkyTemple, so I'll keep this as a draft until I do it (probably tomorrow).

Closes #165.

### DIFF
--- a/skytemple_files/patch/handler/compress_iq_data.py
+++ b/skytemple_files/patch/handler/compress_iq_data.py
@@ -1,0 +1,77 @@
+#  Copyright 2020-2021 Capypara and the SkyTemple Contributors
+#
+#  This file is part of SkyTemple.
+#
+#  SkyTemple is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  SkyTemple is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


skytemple_files/patch/handler/compress_iq_data.py
```diff
@@ -0,0 +1,77 @@
...
+            if config.game_region == GAME_REGION_US:
+                return read_uintle(rom.arm9, OFFSET_US, 4) != ORIGINAL_INSTRUCTION
+            if config.game_region == GAME_REGION_EU:
+                return read_uintle(rom.arm9, OFFSET_EU, 4) != ORIGINAL_INSTRUCTION
+        raise NotImplementedError()
+
+    def apply(self, apply: Callable[[], None], rom: NintendoDSRom, config: Pmd2Data):
+        # Copy the list of skills per group and rewrite it after applying the patch to avoid overwriting
+        # it with the default data included in the patch
+        if self.is_applied(rom, config):
```
theCapypara, line 67: :+1:

theCapypara (approved):

PR #169, opened 2021-09-08
Closes #168: IQ Skill restrictions

### LINKED ISSUE #168: IQ skill restrictions
Split from #18
> Oh and the IQ skill restrictions are also inside arm9.bin (0xa1fe0 -> 0xa2069)
> Each skill has a 2-byte "restriction group" value. If two skills have the same restriction group, they are incompatible and can't be enabled at the same time

US offset: 0xA1A5C

theCapypara:
Closes #168

I have the feeling the data isn't correct yet? I'm not sure where I messed up, can you check?

See also skytemple/skytemple#331

### DIFF
--- a/skytemple_files/hardcoded/iq.py
+++ b/skytemple_files/hardcoded/iq.py
@@ -156,17 +156,10 @@ class HardcodedIq:
         assert (block.end - block.begin) // IQ_SKILL_ENTRY_LEN == 1 + (block_restr.end - block_restr.begin) // IQ_SKILL_RESTR_ENTRY_LEN
         lst = []
         for i in range(0, (block.end - block.begin) // IQ_SKILL_ENTRY_LEN):
-            # Null entry not present in restriction table
-            if i == 0:
-                lst.append(IqSkill(
-                    read_sintle(arm9bin, block.begin + (i * IQ_SKILL_ENTRY_LEN), 4),
-                    None
-                ))
-            else:
-                lst.append(IqSkill(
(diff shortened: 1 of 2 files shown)


skytemple_files/hardcoded/iq.py
```diff
@@ -153,18 +150,40 @@ def set_nectar_gain(value: int, ov29: bytearray, config: Pmd2Data):
...
+            # Null entry not present in restriction table
+            if i == 0:
+                lst.append(IqSkill(
+                    read_sintle(arm9bin, block.begin + (i * IQ_SKILL_ENTRY_LEN), 4),
+                    None
+                ))
+            else:
+                lst.append(IqSkill(
+                    read_sintle(arm9bin, block.begin + (i * IQ_SKILL_ENTRY_LEN), 4),
+                    read_sintle(arm9bin, block.begin + ((i - 1) * IQ_SKILL_RESTR_ENTRY_LEN), 2)
```
End45, line 166:
```suggestion
                    read_sintle(arm9bin, block_restr.begin + ((i - 1) * IQ_SKILL_RESTR_ENTRY_LEN), 2)
```

theCapypara: there we go :)

End45 (changes requested):
I don't know if it's an UI bug, but the restriction values are off by 1. The inital 9999 corresponds to IQ skill #0.

theCapypara:
You are right. The list had one entry less back when I used the wrong table for data which is why I thought it just didn't have a null entry. Fixed.

End45 (approved):
Looks good now

PR #174, opened 2021-09-17
Fix Expand Pokémon List

irdkwia:
This should fix crashes when Female Castform and Cherrim are transforming.


theCapypara:
Thanks!

PR #179, opened 2021-10-11
Update EU ExtractSPCode

ghost:
Wrong path.


marius851000:
It work ! (now, my patch have an issue in that it has no side effect, but at least, spinlock lock the game)

PR #183, opened 2021-11-03
Add DisarmOneRoomMonsterHouses patch

johanngan:
Patch to de-Monster-House-ify the one-room Monster House default layout that appears when the dungeon generation algorithm fails. The speedrunners wanted this.


johanngan:
I don't actually have the dev environment set up, so I haven't tested if this runs properly.

End45 (approved):
Works fine on EU

theCapypara:
Thanks!

PR #187, opened 2021-11-25
Fix Exp Share patch

irdkwia:
Excludes enemies gaining levels by exp. gain.
Enemies never update their exp. points value when they are created so its most of the time invalid.


theCapypara:
Thanks!

PR #204, opened 2022-02-20
Using pmdsky-debug for binaries and symbols

theCapypara:
This switches from using the "Binaries" in "pmd2data.xml" to using the symbol YML files of pmdsky-debug.
@UsernameFodder: Would be great if you could also have a look and add a review (can't add you to reviewers since you aren't a member / contributor).


skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
```diff
@@ -0,0 +1,93 @@
...
+        NA: 0x204EFE0
+        EU: 0x204ECA8
+      length:
+        NA: 0x168
+        EU: 0x168
+      description: Requires EditExtraPokemon patch.
+
+overlay0:
+  data:
+    - name: MAIN_MENU_SONG_ID
```
UsernameFodder, line 51: This one's already in `literals.yml` as `TOP_MENU_MUSIC_ID`.

skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
```diff
@@ -0,0 +1,93 @@
...
+        NA: 0x22BE1A0
+        EU: 0x22BE9B4
+      length:
+        NA: 0x1
+        EU: 0x1
+      description: "Song playing in the main menu. Note: Probably in the middle of a function."
+
+overlay9:
+  data:
+    - name: MAIN_MENU_RETURN_SONG_ID
```
UsernameFodder, line 62: This one's already in `literals.yml` as `TOP_MENU_RETURN_MUSIC_ID`. Let's add the description upstream.

skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
```diff
@@ -0,0 +1,93 @@
...
+        NA: 0x233D900
+        EU: 0x233E080
+      length:
+        NA: 0x1
+        EU: 0x1
+      description: "Song playing in the main menu when returning from the Sky Jukebox. Note: Probably in the middle of a function."
+
+overlay11:
+  data:
+    - name: MONSTER_GROUND_IDLE_ANIM
```
UsernameFodder, line 73: Is this part of a larger function? If we know what the full function does maybe we should add it upstream.

skytemple_files/_resources/ppmdu_config/pmd2data.xml
```diff
@@ -68,592 +68,9 @@
...
-        <Block name="UnknownMenuConfirm" beg="0xD4C" end="0xD64"/>
-        <Block name="DungeonDebugMenu" beg="0xD74" end="0xD9C"/>
-      </Bin>
-      <Bin filepath="overlay/overlay_0035.bin" loadaddress="0x022BD3C0"/>
-      <!-- THIS IS END45'S 'EXTRA SPACE' OVERLAY. IT REQUIRES THE 'ExtraSpace' PATCH -->
-      <Bin filepath="overlay/overlay_0036.bin" loadaddress="0x023A7080"/>
-    </Game>
-
-    <!--Japan-->
-    <Game id="EoS_JP">
```
UsernameFodder, line 583: I was lazy and haven't added the JP values in `pmdsky-debug`, but maybe we should before deleting this file lol

skytemple_files/common/ppmdu_config/pmdsky_debug/util.py
```diff
@@ -0,0 +1,64 @@
...
+            return True
+        if super().__contains__(self._pascal_to_screaming_snake(o)):
+            return True
+        return super().__contains__(self._screaming_snake_to_pascal(o))
+
+    @staticmethod
+    def _pascal_to_screaming_snake(string: str) -> str:
+        return PASCAL_REGEX.sub('_', string).upper()
+
+
```
UsernameFodder, line 61: Why's there two newlines here lol

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,308 @@
...
+        pool[name] = newobj
+        return newobj
+
+
+def _read_symbol(loadaddr: int, symbol_def: dict, typ: Pmd2BinarySymbolType, region: str) -> Optional[Pmd2BinarySymbol]:
+    assert isinstance(loadaddr, int)
+    if 'name' in symbol_def:
+        name = symbol_def['name']
+    else:
+        raise ValueError("Symbol is missing it's name.")
```
UsernameFodder, line 183: its

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,308 @@
...
+    
+    if 'address' in symbol_def:
+        if region in symbol_def['address']:
+            if isinstance(symbol_def['address'][region], list):
+                begin = symbol_def['address'][region][0] - loadaddr
```
UsernameFodder, line 188: This is probably fine, but I wonder if ignoring all but the first value is going to bite at some point in the future...

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,308 @@
...
+            else:
+                begin = symbol_def['address'][region] - loadaddr
+        else:
+            return None
+    else:
+        raise ValueError(f"Symbol {name} is missing an address.")
+    
+    if 'description' in symbol_def:
```
UsernameFodder, line 196: We could simplify this to `description = symbol_def.get('description', "")`

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,308 @@
...
+        description = symbol_def['description']
+    else:
+        description = ""
+    
+    if typ == Pmd2BinarySymbolType.FUNCTION:
+        return Pmd2BinarySymbol(
+            name=name,
+            typ=Pmd2BinarySymbolType.FUNCTION,
+            begin=begin,
+            end=None,
```
UsernameFodder, line 206: Technically there's no reason why a function symbol can't have a length. I just didn't add any because it's a bit easier not to and lengths aren't generally useful for functions. But is there really a need for the `if-else`?

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,308 @@
...
+                if sym is not None:
+                    binary.symbols.append(sym)
+        if 'data' in definition:
+            assert binary.loadaddress is not None
+            for symbol_def in definition['data']:
+                sym = _read_symbol(binary.loadaddress, symbol_def, Pmd2BinarySymbolType.DATA, region)
+                if sym is not None:
+                    binary.symbols.append(sym)
+        if 'description' in definition:
+            if region in definition['description']:
```
UsernameFodder, line 252: Descriptions aren't region-dependent, so I think the description will probably never get set like this.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,308 @@
...
+
+def load_binaries(edition: str) -> List[Pmd2Binary]:
+    version, region = edition.split("_")
+    if version != GAME_VERSION_EOS:
+        raise ValueError("This game version is not supported.")
+
+    binaries: Dict[str, _IncompleteBinary] = {}
+
+    files = glob(os.path.join(SYMBOLS_DIR, '*.yml'))
+    files.sort(key=lambda key: 1 if key.startswith('arm') or key.startswith('overlay') else -1)
```
UsernameFodder, line 299: Am I reading this right? The `sort` method sorts ascending, so assigning 1 to `arm*` and `overlay*` and -1 to everything else is going to put `arm*` and `overlay*` last, not first. I think we want them first so the descriptions make it in properly?

UsernameFodder (changes requested):

.github/workflows/check-pmdsky-debug.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+      - name: Install resymgen
+        run: cargo install resymgen
+      - name: Test
+        run: ~/.cargo/bin/resymgen check --complete-version-list --explicit-versions --in-bounds-symbols --no-function-overlap --nonempty-maps --unique-symbols --data-names SCREAMING_SNAKE_CASE --function-names PascalCase symbols/*.yml
```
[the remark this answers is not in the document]
theCapypara: I didn't check, I didn't want to bother with the possibility of it not being in PATH haha

skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
theCapypara: I completely forgot to check the literals.yml file :clown_face: I just asumed you didn't track immediate values.

skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
theCapypara: This is part of the ChangePokemonGroundAnim patch (#178). I should add that as a description.

skytemple_files/_resources/ppmdu_config/pmd2data.xml
theCapypara: I mean it's still always here in the history haha

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
theCapypara: I will never get this right. :hankey:

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
theCapypara: Yeah I might change it later to allow multiple `begin` values (have the type of `begin` be `Union[int, List[int]]`, but only if needed.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
theCapypara: I can get rid of the if-else, it was mostly useful during writing the function since it would have error'd if there was a data block without a length. But I can just ignore it now if it doesn't exist.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
theCapypara: Thanks, I missed this.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
theCapypara: Oh, you're right. To be honest I just implemented this blindly from memory. I didn't check that `literals.yml` and `ram.yml` also contain the length and addresses (I just asumed they didn't) - That's the main reason I added this, to make sure those are set first. But guess it still works since those two files still contain everything needed.

theCapypara:
I updated the other submodules on accident, but looked through them; should be good. So I'll just keep that in.

theCapypara:
oh also I'll switch to your upstream repo once the PR over there is merged @UsernameFodder. I generally think having the upstreams in here is better than using forks. Since it's a submodule stability is also not affected, since updates would need to be committed here anyway.

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+class MigrationTestCase(unittest.TestCase):
+    """
+    This tests that all blocks and functions that were formely supported by pmd2data.xml are still accessible
+    after migrating to the new pmdsky-debug based solution.
+    """
+    @parameterized.expand((GAME_REGION_EU, GAME_REGION_US))
+    def test_backwards_compatibility(self, region):
+        region = GAME_REGION_EU
```
UsernameFodder, line 226: Was this for temporary testing?

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
```diff
@@ -0,0 +1,244 @@
...
+OLD_ENTRIES = {
+    'arm9.bin': {'CartRemovedImgData': (True, False),
+                 'CompressedIqGroupsSkills': (True, False),
+                 'DebugPrint': (False, True),
+                 'DebugPrint0': (False, True),
+                 'DebugPrint2': (False, True),
+                 'DebugSpecialEpisodeType': (False, False),
+                 'DefaultHeroId': (True, False),
+                 'DefaultPartnerId': (True, False),
+                 #'DungeonData': (False, False),
```
UsernameFodder, line 37: I'm guessing these commented-out entries were never used by SkyTemple? Most of them (maybe not all though) should still be in `pmdsky-debug` under different names I think.

UsernameFodder:
> oh also I'll switch to your upstream repo once the PR over there is merged @UsernameFodder. I generally think having the upstreams in here is better than using forks. Since it's a submodule stability is also not affected, since updates would need to be committed here anyway.

Yeah that's fair, I forgot that submodules always point to a specific commit, so I this is probably an easier way to do things.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,302 @@
...
+                begin = symbol_def['address'][region] - loadaddr
+        else:
+            return None
+    else:
+        raise ValueError(f"Symbol {name} is missing an address.")
+
+    end = None
+    if 'length' in symbol_def:
+        if region in symbol_def['length']:
+            end = begin + symbol_def['length'][region]
```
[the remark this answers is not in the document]
theCapypara: Pretty sure it is. I will run some tests before merging this to confirm all the begin/end addresses still line up.

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
theCapypara: :clown_face:

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
theCapypara: Yeah, they either weren't used or I updated all references to them to the new name manually for one reason or another.

skytemple_files/_resources/ppmdu_config/pmd2data.xml
UsernameFodder: Decided to stop being lazy and added them :P

skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
```diff
@@ -0,0 +1,63 @@
+# Symbols that are too specific to SkyTemple or related mods to be included in pmdsky-debug.
+# Or they are values in the middle of functions.
+
+arm9:
+  functions:
+    - name: DebugPrint2
```
UsernameFodder, line 6: What exactly is this thing? What mod adds this? When I checked the offset it seems like this is a loop within some other function...

UsernameFodder (approved):
This all looks good to me. I'm not familiar with SkyTemple's internal plumbing, but if it passes the tests I assume it should be all good.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
```diff
@@ -0,0 +1,302 @@
...
+        if 'description' in definition:
+            if binary.description == "":
+                binary.description = definition['description']
+
+
+def _build(binaries: Dict[str, _IncompleteBinary]) -> List[Pmd2Binary]:
+    out = []
+    for bin_name, incl_bin in binaries.items():
+        if incl_bin.loadaddress is None:
+            raise ValueError(f"Did not find a load address for {bin_name}.")
```
UsernameFodder, line 251: Err, this isn't going to be called with `EoS_JP` is it? I don't actually know what the JP load addresses/binary sizes are so they aren't in `pmdsky-debug`.

End45:
Lists > IQ crashes:
```
Traceback (most recent call last):
  File "c:/msys64/home/main/skytemple/skytemple/controller/main.py", line 427, in on_view_loaded
    view = controller.get_view()
  File "c:/msys64/home/main/skytemple/skytemple/module/lists/controller/iq.py", line 62, in get_view
    self._init_iq_skills()
  File "c:/msys64/home/main/skytemple/skytemple/module/lists/controller/iq.py", line 300, in _init_iq_skills
    iq_skills = HardcodedIq.get_iq_skills(arm9, static_data)
  File "c:/msys64/home/main/skytemple-files/skytemple_files/hardcoded/iq.py", line 156, in get_iq_skills
    assert (block.end - block.begin) // IQ_SKILL_ENTRY_LEN == 1 + (block_restr.end - block_restr.begin) // IQ_SKILL_RESTR_ENTRY_LEN
AssertionError
```

Lists > Guest Pokémon also crashes (after applying the ASM patch):
```
Traceback (most recent call last):
  File "c:/msys64/home/main/skytemple/skytemple/controller/main.py", line 427, in on_view_loaded
    view = controller.get_view()
  File "c:/msys64/home/main/skytemple/skytemple/module/lists/controller/guest_pokemon.py", line 64, in get_view
    self._fill_guest_pokemon_data()
  File "c:/msys64/home/main/skytemple/skytemple/module/lists/controller/guest_pokemon.py", line 243, in _fill_guest_pokemon_data
    self._get_move_display_name(item.moves[0]), self._get_move_display_name(item.moves[1]),
  File "c:/msys64/home/main/skytemple/skytemple/module/lists/controller/guest_pokemon.py", line 367, in _get_move_display_name
    entry = self.move_entries[i]
IndexError: list index out of range
```

skytemple_files/_resources/skytemple_pmdsky_debug_symbols.yml
theCapypara: I might need to re-check this one. It's possible I mixed something up way back when I investigated the print functions.

skytemple_files/common/ppmdu_config/pmdsky_debug/loader.py
theCapypara: Not at the moment.

theCapypara:
Thanks, will check that.

theCapypara:
@End45 This could also be because of missing changes in the other two repos. I pushed them now:
- https://github.com/SkyTemple/skytemple/pull/386
- https://github.com/SkyTemple/skytemple-ssb-debugger/pull/70

But I'll also diff all the offsets now.

theCapypara:
Opened one PR:
https://github.com/UsernameFodder/pmdsky-debug/pull/8
Both of the examples you gave seem to work now with that and my last commit.

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
```diff
@@ -0,0 +1,243 @@
...
+        for bin_path, symbol_specs in OLD_ENTRIES.items():
+            self.assertIn(bin_path, to_check)
+            bin_symbols = to_check[bin_path].symbols
+            for symbol_name, (has_end, is_function) in symbol_specs.items():
+                self.assertIn(symbol_name, bin_symbols, f"Symbol {symbol_name} in binary {bin_path} must be defined.")
+                symbol = bin_symbols[symbol_name]
+                if has_end:
+                    try:
+                        # noinspection PyStatementEffect,PyUnusedLocal
+                        _x = symbol.end
```
UsernameFodder, line 237: I think I might've inserted padding bytes on a few more data symbols than the one we found with `IQ_SKILL_RESTRICTIONS`. Maybe we change `has_end: bool` to `end: Optional[int]` and explicitly check the lengths here to make sure there are no other surprises.

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
```diff
@@ -0,0 +1,243 @@
+#  Copyright 2020-2022 Capypara and the SkyTemple Contributors
```
UsernameFodder, line 1: In addition to these tests, could it be useful to add a snapshot/diff test for currently defined symbols? That might make it easier to tell what's changed when updating the submodule.

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
theCapypara: I didn't want these tests to include the entire old lookup values, I did write a small test script to check, which can be found here: 
https://gist.github.com/theCapypara/4293b80d1251c0a3b9627528b9d033df

I need to re-check the `SKIP`ed values before merging again, but I think none of these rely on the "correct" end being specified at the moment so it's fine.

skytemple_files/common/ppmdu_config/pmdsky_debug/test/migration_test.py
theCapypara: See above.

theCapypara:
I checked the remaining symbols that don't have a matching end defined now, and their end wasn't used by SkyTemple. So this should be all good now.

End45 (approved):
The crashes are gone, looks like it's working fine.

PR #216, opened 2022-03-23
Rename outdated paths

End45:
Rename a couple of outdated paths that were causing an error when trying to apply the move growth or stat display patches.


theCapypara:
Merci!

PR #226, opened 2022-04-22
Allow editing fixed room item quantities

End45:
The second field in fixed room item data is the quantity. I renamed it so it can be edited.
See also https://github.com/SkyTemple/skytemple/pull/413

Looks like pmdsky-debug got updated too so I guess that goes in the PR as well. Hopefully it won't be an issue.


theCapypara:
Thanks! your branch actually downgraded pmdsky-debug, I'll push the newest commit.

PR #230, opened 2022-05-16
Monster MD: Make unk17 and unk18 unsigned

End45:
These two fields are normally unused (I think. It doesn't seem like the game even has a function to read them.), but the ExpandPokeList patch uses them to store the sprite size and sprite file size of each pokémon (those values are normally stored in a fixed table in arm9.bin). Since the fields are treated as signed integers by SkyTemple, this can cause errors when trying to save pokémon data after applying the patch (for example, see https://discord.com/channels/710190644152369162/787524304932831252/975744066539126814).


theCapypara:
Thanks! Looking at the checks... I sure love how stable Python type checks are.

PR #239, opened 2022-05-26
Update MoveShortcuts patch

End45:
MoveShortcuts has been updated to prevent the player from passing turn by holding A+B while the move list is shown, which would cause a minor graphical glitch.


theCapypara:
Thanks!

PR #244, opened 2022-05-30
Add PitfallTrapTweak patch

Adex-8x:
Adds a small ASM patch that slightly edits the behavior of Pitfall Traps. When the leader activates a Pitfall Trap in ascending dungeons, they now will descend a floor. Also, if the leader steps on a Pitfall Trap in 1F, they will restart 1F with a different seed.

I don't have a dev environment set up yet, so aside from having tried this out with patch packages, I'm unable to test if anything breaks while doing all of this.

### DIFF
--- a/skytemple_files/_resources/ppmdu_config/pmd2data.xml
+++ b/skytemple_files/_resources/ppmdu_config/pmd2data.xml
@@ -862,6 +862,15 @@
           </Replace>
         </SimplePatch>
 
+        <Patch id="PitfallTrapTweak">
+          <OpenBin filepath="overlay/overlay_0029.bin">
+	  	<Include filename ="adex_asm_mods/pitfall_trap_tweak/selector_overlay29.asm"/>
+	  </OpenBin>
+	<OpenBin filepath="overlay/overlay_0036.bin">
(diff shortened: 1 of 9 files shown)


skytemple_files/_resources/ppmdu_config/pmd2data.xml
```diff
@@ -862,6 +862,15 @@
...
         </SimplePatch>
 
+        <Patch id="PitfallTrapTweak">
+          <OpenBin filepath="overlay/overlay_0029.bin">
+	  	<Include filename ="adex_asm_mods/pitfall_trap_tweak/selector_overlay29.asm"/>
+	  </OpenBin>
+	<OpenBin filepath="overlay/overlay_0036.bin">
+	  	<Include filename ="adex_asm_mods/pitfall_trap_tweak/selector_overlay36.asm"/>
+	  </OpenBin>
+        </Patch>
```
End45, line 872: Borked indentation

End45 (changes requested):
Tested on an EU ROM, everything seems to be working fine.

theCapypara:
Thank you! :tada:

theCapypara:
The code format and Pylint check failing is normal atm, don't worry about those.

PR #246, opened 2022-06-05
Update pmdsky-debug to `c0fc599`

UsernameFodder:
The `--no-function-overlap` option for `resymgen check` was renamed to `--no-overlap`, so check-pmdsky-debug.yml needs to be updated accordingly.

I'm not sure when this submodule was last updated...I'm hoping there were no other breaking changes introduced since then.


theCapypara:
Thanks, tests run through and at a glance at the symbol changes everything seems fine.

PR #250, opened 2022-06-22
Support to download and import SpriteCollab assets

theCapypara:
This adds support to download and import sprites and portraits from spritecollab-srv servers to SkyTemple Files.


skytemple_files/hardcoded/ground_dungeon_tilesets.py
```diff
@@ -116,7 +116,7 @@ def resolve_mapping_for_level(
     mappa_idx = dungeons[dungeon_id].mappa_index
     start_offset = dungeons[dungeon_id].start_after
     length = dungeons[dungeon_id].number_floors
-    floor_id = cast(u8, min(length - 1, floor_id))  # type: ignore
+    floor_id = cast(u8, min(length - 1, floor_id - 1))  # type: ignore
```
theCapypara, line 119: Unrelated other bug I found.

PR #261, opened 2022-07-08
Add FixNo$GbaSaves patch

End45:
Adds a patch that fixes saving in the No$GBA emulator, which allows playing the game without having to enable a cheat every time the emulator is opened.

I haven't been able to test this in SkyTemple since the latest dev version freezes when trying to load the ROM. The ASM patch itself is tested on both EU and US.


End45:
Alright, SkyTemple ended up loading the ROM, I just had to wait a bit longer.
The patch has now been tested on SkyTemple, everything works fine.

theCapypara:
Thanks!

PR #262, opened 2022-07-09
Fix obj_table breaking weather on EU rom

marius851000:
I figured that the screen turned black except for dialogue in a cutscene of an hack I'm working on when applying this patch. After a bit of reverse-engineering, I found it was due to the padding overwriting some table that is used to check the ID of the of the background (in bg_list.dat) to load for the weather layer with 0xCC.

This patch fix some offset. The change itself was made by the original patch creator. (I got permission from them to upstream it).

It quickly tested it (as in, it didn't worked before, it work now. Didn't tried to add new object thought).


theCapypara:
Thanks!!

PR #271, opened 2022-07-22
fixed trap name position in EU

marius851000:
It seems there is a trap description at the end of the list of trap name


theCapypara:
Thanks!

PR #275, opened 2022-07-30
Suggestion: Make enum args checks more generic

irdkwia:
All enum arguments checks are exactly the same (save the const__by_id attribute in script_data), so replacing it with a single check using a dictionary mapping each enum to its script_data attribute should be better.
Would need to change rust implementation and rust interface files at: 
https://github.com/SkyTemple/skytemple-rust/blob/1852c5eb97bc6d177cc3d6f58fa338ff3514a48f/skytemple_rust/st_ssb.pyi#L52
https://github.com/SkyTemple/skytemple-rust/blob/1852c5eb97bc6d177cc3d6f58fa338ff3514a48f/skytemple_rust/st_ssb.pyi#L56

### DIFF
--- a/skytemple_files/common/ppmdu_config/script_data.py
+++ b/skytemple_files/common/ppmdu_config/script_data.py
@@ -458,8 +458,8 @@ class Pmd2ScriptData(AutoString):
         self._common_routine_info = value
 
     @property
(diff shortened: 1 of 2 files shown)


theCapypara:
Thanks!

PR #277, opened 2022-08-12
Add BoldText and ChangeTextSound patches

Adex-8x:
Adds two ASM patches that implement new text tags to be used in strings.
- BoldText adds the [BS] and [BR] text tags, where any text encased within them will be bolded.
- ChangeTextSound adds the [TS:X] (where X is an integer) and [TR] text tags. [TS:X] sets the textbox to use the Xth sound effect ID, while [TR] reverts the sound to its default.

### DIFF
--- a/skytemple_files/_resources/patches/asm_patches/adex_asm_mods/bold_text/common/bold_arm9.asm
+++ b/skytemple_files/_resources/patches/asm_patches/adex_asm_mods/bold_text/common/bold_arm9.asm
@@ -1,0 +1,9 @@
+.org NoBARFound
+.area 0x4
+	beq HookBLetter
+.endarea
+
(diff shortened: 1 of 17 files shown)


theCapypara:
oh cool, thanks! Can you make sure the code format check runs through? You just need to install black for that and run it in the directory of the repo:
https://github.com/psf/black

Adex-8x:
There's a file unrelated to my patches (`skytemple_files/script/ssb/model.py`) that fails the format check, did you want me to upload that as well?

theCapypara:
Oh weird, probably from Irdkwia's PR and I missed it. Would be great yeah

theCapypara:
Thanks!

PR #278, opened 2022-08-21
Add string interning for const SSB strings

tech-ticks:
Interns (de-duplicates) strings during SSB compilation to save space (there are no ulterior motives).

This is completely untested since I couldn't be bothered to update and fix my development setup right now. I'm not deep enough into SSB internals to tell if this change could break anything.


theCapypara:
Thanks, I'll test this later.

theCapypara (approved):

PR #279, opened 2022-08-23
Updated CTC to 1.2.4

MkTr:
Players are now prevented from changing team tactics during manual mode. For some reason, doing so completely destroys turn order and leadership status. 
(ignore the comment about this being a duplicate, I forgot to press the create pull request button last night)

### DIFF
--- a/skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
+++ b/skytemple_files/_resources/patches/asm_patches/cipnit_asm_mods/completeteamcontrol.asm
@@ -1,4 +1,4 @@
-; PMD EOS - Complete Team Control Code v1.2.3
+; PMD EOS - Complete Team Control Code v1.2.4
 ; Made by Cipnit
 ; https://www.pokecommunity.com/showthread.php?t=437108
 ; Build this file using armips: https://github.com/Kingcom/armips
@@ -183,6 +183,7 @@
 	push r0
 	cmp r1,34h		;farewell decision
 	cmpne r1,3Bh	;change leader decision
+	cmpne r1,1Ah	;tactics decision, screws with turn order for some reason
 	bne @@paranoiaagent_nvm
 	ldr r0,=@ManualModeOn
 	ldrb r0,[r0]


End45:
> For some reason, doing so completely destroys turn order and leadership status.

How odd. Do other actions that involve interacting with a menu (like enabling or disabling moves) also have this problem?

MkTr:
> How odd. Do other actions that involve interacting with a menu (like enabling or disabling moves) also have this problem?

Enabling/disabling moves are fine - I doubt any common actions would be a problem since the hack has been available for years, and this likely went undiscovered because setting tactics during manual mode is never needed. What made it more elusive is the fact that only a single tactic triggers the bug.
When the player presses start to change modes, my code checks if the first pokemon's action variable is equal to 0 (line 107 in the source code). If it is, that means it's the first pokemon's turn, and changing modes is allowed, as the action variable is set to 0 at the start of a leader's turn (before you choose its decision). At any other time, and for AI-controlled pokemon, the action variable is ALWAYS nonzero, as doing anything, even skipping your turn, sets it to a nonzero number. 
For some reason, setting a pokemon's tactics also sets its action variable. This is completely fine for most of the tactics, which set it to 1 or 2, but for whatever reason, choosing "wait there" sets the action variable to 0, which makes it possible to change modes during a partner's turn. So yeah, any option which can change another pokemon's action variable to 0 will trigger this bug - I haven't found any others. 
I thought there was another bug where changing a pokemon's tactic screwed with the turn order (during manual mode, mon 1 moves, mon 2 changes mon 3's tactics then moves, mon 3 moves, then mon 3 has another AI-controlled turn before mon1 moved), but it could have been because I accidentally tested it on an old version where partners randomly getting extra turns was an issue - I'm not able to reproduce the bug in v1.2.3. 

Also, I'm not sure what to do about the "some checks were not successful" error messages github's giving (Build, test and publish / Runs tests with real ROM). The failed checks all have "ERROR:   py38: commands failed", with py38, py39, and py310. I've tested it myself, it definitely runs on a real rom, so I've got no idea what the issue is.

theCapypara:
Hi, thanks a lot for the update! I'll re-run the tests, they should be fine.

PR #284, opened 2022-09-15
Add AntiSoftlock patch

End45:
Adds a patch that allows the user to press A+B+X+Y during cutscenes to unlock all locks at once. This can be used to fix many of the softlocks caused by changing the starters.

This patch could also be added as a default patch to the randomizer, since these kind of softlocks are common there.


theCapypara:
Cool, thanks!

PR #312, opened 2022-12-30
Fix ExtraSpace corruption

### LINKED ISSUE #311: Latest change to the ExtraSpace handler corrupts the ROM
The change made to the ExtraSpace patch handler in #307 fixes how the patch is applied but creates a different problem: Instead of resulting in an empty overlay36, now `BACK/expback.bgp` is emptied, which causes a crash whenever that file is displayed in-game (for example, on the scene where you need to choose your team name).

theCapypara:
Fixes #311

### DIFF
--- a/skytemple_files/common/util.py
+++ b/skytemple_files/common/util.py
@@ -605,8 +605,10 @@ def set_binary_in_rom(rom: NintendoDSRom, binary: SectionProtocol, data: bytes) 
     raise ValueError(f(_("Binary {binary.name} not found.")))
 
 
-def is_binary_in_rom(rom: NintendoDSRom, binary: SectionProtocol) -> bool:
+def is_binary_in_rom(rom: NintendoDSRom, binary: Optional[SectionProtocol]) -> bool:
     """Returns true if the specified binary is present in the rom"""
+    if binary is None:
+        return False
     try:
         get_binary_from_rom(rom, binary)
         return True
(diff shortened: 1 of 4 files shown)


End45:
Looks like it works. The main thing I don't like about how this is currently set up is that if someone applies a patch that tries to write to overlay 36 and ExtraSpace isn't applied (which can happen if they forget to mark the patch as dependant on ExtraSpace, for example), armips will be able to write to the overlay on the tmp folder, but changes won't get saved to the ROM since the overlay isn't on the table. It would be better if trying to do this threw an error instead of hiding the problem.

This could be implemented with 3 changes:

1) Change https://github.com/SkyTemple/skytemple-files/blob/034fcba141b32e80d1d6c6dcbd7d5882195356c0/skytemple_files/patch/patches.py#L344 so overlay 36 is only added to the list if it's already on the ROM or ExtraSpace is being applied.

2) Change
```
except ValueError as err:
    if binary_name == "overlay36":
        # SPECIAL CASE for ExtraSpace patch.
```
with `if binary_name == "overlay36" and <ExtraSpace is being applied>` so it throws an error if the file is missing in other circumstances.

3) Get rid of https://github.com/SkyTemple/skytemple-files/blob/034fcba141b32e80d1d6c6dcbd7d5882195356c0/skytemple_files/patch/arm_patcher.py#L225, since that shouldn't happen anymore under any situation (if ExtraSpace isn't applied, then overlay 36 won't be on the list of files to save and if we're applying it or it's already applied, it will be on the overlay list already so `set_binary_in_rom()` won't fail).

theCapypara:
Can you make these changes or suggest them using GitHubs built in editor? I sadly don't have time for a while to look further into this.

End45:
I guess I could try. I think I'll have to open another PR though, since I don't think I can push on this one.

End45 (approved):
Alright, everything looks good now!

PR #314, opened 2022-12-30
Suggestion for #312

End45:
Implements my suggestion to ensure that trying to patch overlay 36 when ExtraSpace isn't applied results in an error.

### DIFF
--- a/skytemple_files/common/util.py
+++ b/skytemple_files/common/util.py
@@ -605,6 +605,15 @@ def set_binary_in_rom(rom: NintendoDSRom, binary: SectionProtocol, data: bytes) 
     raise ValueError(f(_("Binary {binary.name} not found.")))
 
 
+def is_binary_in_rom(rom: NintendoDSRom, binary: SectionProtocol) -> bool:
+    """Returns true if the specified binary is present in the rom"""
+    try:
+        get_binary_from_rom(rom, binary)
+        return True
+    except ValueError:
+        return False
+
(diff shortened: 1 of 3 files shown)


End45:
I don't know enough Python to understand why there's errors about incompatible types. I tested everything and didn't get any errors at runtime.

Also, once again I don't understand what the formatter wants from me. It's not very helpful if it doesn't tell you what the problem is exactly.

theCapypara:
Thanks!

> I don't know enough Python to understand why there's errors about incompatible types. I tested everything and didn't get any errors at runtime.

```
skytemple_files/patch/patches.py:343: error: Argument 2 to "is_binary_in_rom" has incompatible type "Optional[SectionProtocol[ExtraOverlay36FunctionsProtocol, ExtraOverlay36DataProtocol, int]]"; expected "SectionProtocol[Any, Any, Any]"
```
overlay36's symbol definition (`self._config.extra_bin_sections.overlay36`) may be None, this can happen for example with the JP ROM. `is_binary_in_rom` requires it to never be None. I'll add a None check there.

```
skytemple_files/patch/patches.py:348: error: Argument 2 to "apply" of "ArmPatcher" has incompatible type "Dict[str, object]"; expected "Dict[str, SectionProtocol[Any, Any, Any]]"
```
The Dict now contains an object that may be None (the overlay), which confuses mypy. I'll fix this a well.

PR #329, opened 2023-02-03
Add BetterEnemyEvolution patch

End45:
Adds a patch that allows adding new rules to enemy evolution, such as updating the stats or moves of the evolved enemy or making the evolution trigger even if the target revives.

Note: This will fail the format check since https://github.com/SkyTemple/skytemple-files/commit/90ea063b2559a55333b43b722ed45c6893493ff0 failed it too.


theCapypara (changes requested):
Hi and thanks! Please make sure all checks run through and then re-request me as reviewer.

End45:
I don't really get why trying to update the submodule to the latest commit results in a conflict, but this PR requires the submodule to be on commit https://github.com/End45/EoS-asm-hacks/commit/7ff71f26089d7022d3067bb6e5170fc5c73a35e7.

End45:
Uh... is that fail my fault?

theCapypara:
No, that is a network issue with the sprite repository test.

PR #332, opened 2023-02-09
Update FixNoCashGbaSaves description

End45:
Adds a warning to the patch description, since it seems like this patch breaks saves on MelonDS instead.


theCapypara (changes requested):
Hi and thanks! Please make sure all checks run through and then re-request me as reviewer.

theCapypara (approved):

PR #342, opened 2023-02-23
Fix softlock caused by AntiSoftlock patch

End45:
How ironic. It could prevent others from softlocking, but not itself.

This fixes a bug that causes a softlock in real hardware and some emulators when A+B is pressed.
Since other PRs that update my patches repo haven't been merged yet (#329, #339), those updates are included here as well.


theCapypara (changes requested):
Hi and thanks! Please rebase and then re-request me as reviewer.

theCapypara (approved):

PR #353, opened 2023-04-09
Add NoWeatherStop patch

End45:
Adds a patch that changes weather damage messages so they don't make the player stop running.
There's still a small delay when the message shows up, but it's no longer necessary to press the button again to keep running.


theCapypara:
Thanks!

PR #362, opened 2023-05-07
Update UnusedDungeonChancePatch

End45:
I've added a fix that prevents wall maze rooms from spawning secondary terrain on them, which could cause some tiles to be unreachable, potentially softlocking the player if the stairs happened to be there.

I also moved the patch from its own separate file into my repo.


theCapypara:
Thanks!

PR #374, opened 2023-07-01
Fix UnusedDungeonChance crash

End45:
Fixes a crash introduced in UnusedDungeonChance v0.1.1 that could happen when generating a wall maze room.

### DIFF
--- a/skytemple_files/patch/handler/unused_dungeon_chance.py
+++ b/skytemple_files/patch/handler/unused_dungeon_chance.py
@@ -36,7 +36,7 @@ OFFSET_EU = 0x642C8
 OFFSET_US = 0x64024
 
 
-class UnusedDungeonChancePatch(AbstractPatchHandler):
+class UnusedDungeonChancePatch(AbstractPatchHandler, DependantPatch):
     @property
     def name(self) -> str:
         return "UnusedDungeonChance"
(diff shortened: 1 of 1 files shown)


theCapypara:
@End45 Thanks! Please rebase your changes on `version/1.5.x`. I was going to do it, but then I saw it's the master branch of your fork, so that's probably not a good idea. You may need to re-open the PR if you want to use another branch.

theCapypara:
Also I see in the source, that this needs the ExtraSpace patch now? In that case you need to mark it as dependant on that.

End45:
> @End45 Thanks! Please rebase your changes on `version/1.5.x`. I was going to do it, but then I saw it's the master branch of your fork, so that's probably not a good idea. You may need to re-open the PR if you want to use another branch.

Did it on master because I expected it was going to be an easy merge. Should have done it properly lol. I'll fix it.

> Also I see in the source, that this needs the ExtraSpace patch now? In that case you need to mark it as dependant on that.

Oops, right. I was going to do it and forgot. Thanks.

theCapypara:
@End45 Can you make sure the type checks run through? At least one of them will probably actually cause runtime issues.

End45:
Right, forgot about the imports

Issue #395, opened 2023-08-01
Add Expanded Abilities


ShiningDeneb:
EoS allows ability numbers up to 254. I've already written a public patch that allows expanding the ability strings, thus allowing new abilities to be added to the game. However editing monster.md to add those abilities causes the game to break SkyTemple, that's why I'm opening this issue. I have an updated version of data/md/protocol.py available but I don't seem to have the requirement to push my changes to the repo, that's why I'm opening this issue.

theCapypara:
Hi! You can submit your changes as a pull request.

theCapypara:
Oh wait, you did. Sorry I somehow didn't see that.

PR #396, opened 2023-08-01
Add Expanded Abilities to the Ability enum

ShiningDeneb:
Solves this issue https://github.com/SkyTemple/skytemple-files/issues/395 A better implementation could be provided later using the actual ability strings instead of using enums

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
 # mypy: ignore-errors
-__version__ = '1.6.0a1'
+__version__ = '1.6.0a2'
 import os
 
 from setuptools import setup, find_packages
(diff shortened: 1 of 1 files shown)


theCapypara:
@ShiningDeneb hi! I added a commit to make the localization a bit less tedious. Is this OK for you?
Also if you have a patch to make use of the additional ability slots and maybe even a possible idea for an UI to edit them, that would be much appreciated as a contribution!
But in general thanks for contributing this PR!

theCapypara:
Sorry I just pushed my changed wrong and accidentally pushed the current master onto your master. I reopened a PR with your changes (they are still attributed to you) here: https://github.com/SkyTemple/skytemple-files/pull/397

theCapypara:
If you can intergrate the script into skytemple-files and add the patch like the rest (maybe @End45 can help with that part) I can do the UI changes required.

PR #403, opened 2023-08-28
Update FixMemorySoftlock to v2

End45:
Updates the FixMemorySoftlock patch to a new version. The patch will now prevent crashes caused by the game running out of memory by replacing the sprite that can't be loaded with a default preloaded sprite. This should fix all the cutscene crashes caused by memory errors.

After this gets merged, the randomizer should be updated as well. The patch should probably applied by default.


theCapypara:
Thanks!
Need to check again, but I'm pretty sure the Randomizer is configured to already apply this by default. So no explicit change in the Randomizer is needed, just a version bump of skytemple-files.

Issue #235, opened 2022-05-20
Japanese Symbols


theCapypara:
The following pmdsky-debug symbols would need to be defined for the JP ROM for SkyTemple to support it:

arm9:
- [x] DebugPrint0
- [x] DebugPrint
- [x] SaveScriptVariableValue
- [x] SaveScriptVariableValueAtIndex
- [x] GetDebugFlag
- [x] GetDebugLogFlag
- [x] SetDebugFlag
- [x] SetDebugLogFlag
- [x] SCRIPT_VARS
- [x] SCRIPT_VARS_LOCALS
- [x] CART_REMOVED_IMG_DATA
- [x] DEFAULT_PARTNER_ID
- [x] DEFAULT_HERO_ID
- [x] PARTNER_START_LEVEL
- [x] HERO_START_LEVEL
- [x] SPECIAL_EPISODE_MAIN_CHARACTERS
- [x] DUNGEON_DATA_LIST
- [x] DUNGEON_RESTRICTIONS
- [x] SECONDARY_TERRAIN_TYPES
- [x] MAP_MARKER_PLACEMENTS
- [x] GUEST_MONSTER_DATA
- [x] MIN_IQ_EXCLUSIVE_MOVE_USER
- [x] MIN_IQ_ITEM_MASTER
- [x] IQ_GUMMI_GAIN_TABLE
- [x] GUMMI_BELLY_RESTORE_TABLE
- [x] WONDER_GUMMI_IQ_GAIN
- [x] JUICE_BAR_NECTAR_IQ_GAIN
- [x] IQ_SKILLS
- [x] IQ_SKILL_RESTRICTIONS
- [x] IQ_GROUP_SKILLS
- [x] MONSTER_SPRITE_DATA
- [x] RANK_UP_TABLE
- [x] LOADED_OVERLAY_GROUP_1
- [x] GAME_STATE_VALUES
- [x] LANGUAGE_INFO_DATA
- [x] GAME_MODE
- [x] NOTIFY_NOTE

ram:
- [x] SCRIPT_VARS_VALUES
- [x] DEBUG_SPECIAL_EPISODE_NUMBER

arm9 extra (introduced via patch, definitions in skytemple-files):
- [x] COMPRESSED_IQ_GROUP_SKILLS
- [x] EXTRA_DUNGEON_DATA
- [x] GUEST_MONSTER_DATA2

overlay0:
- [x] TOP_MENU_MUSIC_ID

overlay1:
[...]

PR #406, opened 2023-09-04
Some extra JP symbols

Adex-8x:
Adds some necessary JP "extra" symbols in arm9 and overlay11, as listed in https://github.com/SkyTemple/skytemple-files/issues/235.


theCapypara:
Thank you!

PR #427, opened 2023-09-20
Attempt to fix JP string blocks

theCapypara:
@Laioxy @Adex-8x I can't add you as reviewers since you are not part of the SkyTemple GitHub organization. But can you have a look?

I tried to fix 
- https://github.com/SkyTemple/skytemple/issues/603
- https://github.com/SkyTemple/skytemple/issues/618

I tried this by moving the IDs for "Move Descriptions" and "IQ Skills Names" down/up one entry. This left a lot of overlap for the other string groups. I think the list is just in general still kind of wrong. What do you think?


theCapypara:
Oh, I probably did it wrong... for IQ Names it should be moved up I guess...

theCapypara:
Note: The UI in SkyTemple starts counting at 1. So 1 in SkyTemple is 0 in reality.

theCapypara:
The "end" numbers are always the ID of the last string that is still part of this group (= inclusive)

theCapypara:
grafik
grafik

Adex-8x:
Both look good at a glance, yeah

Laioxy:
Both look good! I checked too.

Issue #405, opened 2023-08-31
Base Patch support for JP ROM


theCapypara:
In order for SkyTemple and the Randomizer to properly support the JP ROM, the following patches, XML definitions and Python patch handlers need to be updated to support the JP ROM:

- [ ] ActorAndLevelLoader
- [ ] ExtractHardcodedItemLists
- [ ] EditExtraPokemon
- [ ] CompressIQData
- [ ] ProvideATUPXSupport
- [ ] ExtraSpace
- [ ] AntiSoftlock
- [ ] MoveShortcuts
- [ ] UnusedDungeonChance
- [ ] CompleteTeamControl
- [ ] FarOffPalOverdrive
- [ ] PartnersTriggerHiddenTraps
- [ ] ReduceJumpcutPauseTime
- [ ] NoWeatherStop
- [ ] ChooseStarter
- [ ] SkipQuiz
- [ ] FixMemorySoftlock
- [ ] DisarmOneRoomMonsterHouses

This is the base list needed for the Randomizer. Of course more patches would be better, especially:

- [ ] ExtractItemCode
- [ ] ExtractMoveCode
- [ ] ExtractObjectTable
- [ ] ExtractSPCode
- [ ] ExtractDungeonData
- [ ] ExtractBarItemList
- [ ] ExtractAnimData
- [ ] ExtractHardcodedItemLists

PR #431, opened 2023-09-26
Japanese compatibility for most ASM Patches + Patch Updates

Adex-8x:
Adds Japanese ROM compatibility for most ASM Patches! Should close https://github.com/SkyTemple/skytemple-files/issues/405.
Additionally, various ASM Patches have received updates:

- `AddTypes` now has a patch parameter to determine the Conversion 2 type matchup against Fairy.
- `AddTypes` has a "bugfix" that reverts the "Normal -> Ghost" and "Fighting -> Ghost" matchups to be effective instead of immune. This is actually how it is in the base game, for some reason.
- `CompleteTeamControl` now uses Text Strings instead of an ASCII string baked into the patch, to allow for translations.
- `PitfallTrapTweak` now hides the UI when going back a floor.

As of now, the following patches do not yet have JP ROM compatibility:

- `ChangeMoveStatsDisplay`: Crashes when attempting to view a move's description in ground and dungeon mode. This also seems to softlock in the EU version as well, so it may be an underlying issue with the patch in general.
   - See the "TODO STAT_DISP:" comments for areas of concern.
- `ExpandPokeList`: Untested due to an unknown issue (SkyTemple hangs indefinitely)! I have various concerns with this patch in regards to dungeon mode, so having others take a look at this is strongly encouraged! ~~I'm suffering so much with this one please help someone anyone~~
   - See the "TODO EXPAND_POKE_LIST:" comments for areas of concern.
- `ExternalizeMappaFile`: While this patch does not crash on hardware, it seems to read certain attributes incorrectly. For example, trying to place Pitfall Traps in a dungeon after applying this patch actually placed Chestnut Traps.
- `MoveGrowth`: Relies on `ChangeMoveStatsDisplay`.

I apologize for not fully supporting them in this PR, but I at least wanted to add what I have so far, and get other opinions (be it now or in the future) as to what may be causing these last few issues.

Additionally, I would suggest that @Laioxy take a look at `fairy_gummies.py` to see if the Japanese translation of the Fairy Gummi is accurate.

Testing of all of these patches on a platform that _isn't_ DeSmuME is strongly encouraged, as although I have tested them myself, there's always a chance I missed something.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,6 @@
 [submodule "skytemple_files/_resources/patches/asm_patches/pmd2_asm_mods"]
 	path = skytemple_files/_resources/patches/asm_patches/pmd2_asm_mods
-	url = https://github.com/SkyTemple/pmd2_asm_mods.git
-	branch = dev
+	url = https://github.com/PsyCommando/pmd2_asm_mods.git
 [submodule "skytemple_files/_resources/patches/asm_patches/end_asm_mods"]
 	path = skytemple_files/_resources/patches/asm_patches/end_asm_mods
 	url = https://github.com/End45/EoS-asm-hacks.git
--- a/skytemple_files/_resources/patches/asm_patches/adex_asm_mods/bold_text/jp/offsets.asm
+++ b/skytemple_files/_resources/patches/asm_patches/adex_asm_mods/bold_text/jp/offsets.asm
@@ -1,0 +1,10 @@
+.relativeinclude on
+.nds
+.arm
+
+.definelabel NoBARFound, 0x02021030
+.definelabel CallDisplayChar, 0x02021BF4 
+.definelabel TagCheck, 0x02020918
+.definelabel TagCodeError, 0x02021A60
+.definelabel AfterTagIsFound, 0x02021AF0
+.definelabel StackThing, 0x70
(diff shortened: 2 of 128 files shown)


Laioxy:
I have checked the Japanese translation of Fairy Gummi. Rather than using the type name "Fairy" as it is, I think it would be more appropriate to use a color name that evokes the appearance and type.
For example, "うすももグミ"(Usumomo Gumi). It means light pink.
I really wanted to name it "ももいろグミ" but gave up because it is already used for poison type gummies.

Adex-8x:
Ah I see, thank you for checking it!
Maybe a name like "うすももグミ" would be more fitting, yeah. We'll see what Para thinks as well
My only worry is that it wouldn’t be consistent with other languages’ translations

theCapypara:
First, thank you a lot! I'll check everything later. 

How is that translation wrong with the other language translations? Aren't gummies always named after color? If the Fairy Gummi is currently called that in the other languages then that's probably wrong.

Adex-8x:
Gummies are usually named after their color, yeah, but the [current translations](https://github.com/SkyTemple/skytemple-files/blob/fd26f876f48e42b69f08702e726669c36ea7826c/skytemple_files/patch/handler/fairy_gummies.py#L55-L59) opt to literally go for "Fairy Gummi".

theCapypara:
This PR also includes https://github.com/PsyCommando/pmd2_asm_mods/pull/5

theCapypara (approved):
LGTM. I haven't tested all patches but generally this seems good. If okay for everyone I would merge.

Adex-8x:
All fine with me! Whenever I’ve got more time I’ll be able to look into the last four patches that are giving us trouble.

Laioxy:
You may disregard my comments regarding translation! I was talking about something like naming conventions 💦
If a name like "Fairy Gummi" has already been established, then that is fine. By the way, the Japanese translation of "フェアリーグミ"(Fairy Gummi) itself is correct.

theCapypara:
I mean personally I still think we should maybe change it for all languages to be a color. But maybe that can be an extra Pull Request.

PR #434, opened 2023-10-03
Python Upgrade: 3.9 - 3.12

theCapypara:


theCapypara:
SpriteCollab browser is not working: https://github.com/graphql-python/gql/issues/447

Issue #103, opened 2021-04-08
Importing some sprites: Some colors are transparent


theCapypara:
Seems to affect the Glaceon sprites (see SpriteBot) - maybe the default transparent color?

---
Beep, boop, I'm [a bot](https://github.com/Parakoopa/IssueBot)! This issue was created by `@Parakoopa` in `#rom-hacking-support`.

PR #445, opened 2023-11-17
fix glaceon transparency bug

scbroede:
should fix https://github.com/SkyTemple/skytemple-files/issues/103

essentially, running quantize on an RGBA image can reduce the amount of colours in the palette even below the number passed into it (see https://github.com/python-pillow/Pillow/issues/5204 for further details). this poses a problem for glaceon specifically because one of its colours (0, 127, 159) is *really* close to the one defined as transparent.

since the actual transparent pixels have already been replaced with "transparent" pixels at this point, we can just convert it to RGB before running quantize to avoid this


theCapypara:
Thanks!

PR #446, opened 2023-11-17
fix backup transparent colours

scbroede:
this can potentially loop infinitely because `transparent[3] - 1` is always 254. i think this was meant to use the previous colour's blue value, not its alpha value


theCapypara:
Thanks!

PR #447, opened 2023-11-20
Add various ASM patches

Adex-8x:
Adds three ASM patches:
- `AddKeyCheck`
  - Disallows a party member from using a Key (Item ID 182) in the dungeon menu unless they're standing right below a Key Door.
- `ChangePortraitMidText`
  - Implements the text tag `[FACE:X]`, where "X" is the new portrait emotion ID to change the currently-loaded portrait to. This works in ground and dungeon mode, and will default to the default (FACE_NORMAL) portrait if the specified portrait emotion cannot be found.
  - ~~@AlliterativeAlias are you proud of me~~
- `DisplayScriptVariable`
  - Implements the text tag `[var:x:y]`, where "x" is a script variable ID and "y" (optional; will default to 0 if not present) is an index. The text tag will be replaced with the value of the specified script variable.


theCapypara:
Thanks! Those will be released with 1.6.1.

PR #448, opened 2023-11-21
fix double flipping metaframepieces

scbroede:
if the previous piece is already flipped, then hflip should be set to false on the new piece to flip it again


marius851000:
So that’s why those sprites where wrong? Good to know it wasn’t my fault. Thanks. (to be clear, I didn’t tested that, but it looks good to me)

theCapypara:
Thanks!

PR #456, opened 2023-12-22
Added names for unk parameters from wiki

Chesyon:
Updated many of the parameters named "unk", using parameter names from the SkyTemple wiki: https://wiki.skytemple.org/index.php/List_of_Opcodes


theCapypara:
Thanks!

PR #458, opened 2023-12-24
More updates to parameters from wiki

Chesyon:
All of the parameters here should now be fully consistent with https://wiki.skytemple.org/index.php/List_of_Opcodes


theCapypara:
Hi, your changes cause some issues with decompiling/compiling scripts from the game. Please check those. If you need any help, let me know.

theCapypara:
@Chesyon Particularly "Addr" is not a valid argument type.

Chesyon:
Sorry about that! It should hopefully be fixed now. I'll update the wiki accordingly, once the checks finish, assuming that everything works properly this time.

theCapypara:
For the wiki Addr is probably fine. Any opcode with a jump to an adress isn't really used anyway since all of them have language syntax.

theCapypara:
Thanks!

PR #461, opened 2023-12-27
Allow dependabot and forks to publish test results

theCapypara:


theCapypara:
Can apparently only be tested if merged to master...

PR #462, opened 2023-12-29
Change enemy density to unsigned byte

End45:
Hopefully I didn't miss anything.
This breaks compatibility with existing floor XMLs that contain negative spawn weights. It's not a commonly used feature so I don't know if it's worth doing something about that.

PRs in other repos: https://github.com/SkyTemple/skytemple/pull/687, https://github.com/SkyTemple/skytemple-randomizer/pull/166


theCapypara:
The tests fail, because this also requires a change in skytemple-rust:
https://github.com/SkyTemple/skytemple-rust/blob/master/src/st_mappa_bin/layout.rs

End45:
As discussed in Discord, creating an ASM patch that re-enables the intended functionality seems like a better idea, so I'm going to close this and will open a PR with the patch once I make it.

Issue #463, opened 2023-12-29
ExternalizeWazaFile + DungeonInterrupt crash on hardware


Adex-8x:
Originally reported at https://discord.com/channels/710190644152369162/1190254594765369364

If `ExternalizeWazaFile` is applied, attempting to view the Chimecho Assembly menu will cause the game to hard crash. This is confirmed to happen on NA and most likely happens on EU and JP as well.
<img width="174" alt="Screenshot 2023-12-29 at 3 44 57 PM" src="https://github.com/SkyTemple/skytemple-files/assets/91418697/ed5fc902-dc46-4a8a-b4bf-4aa6736d1b57">
Based on the error screen, it looks like the problematic line is around [here](https://github.com/SkyTemple/skytemple-files/blob/3420cbc995fec5b9f9494b7d88a20ffebd43dc1c/skytemple_files/_resources/patches/asm_patches/anonymous_asm_mods/fs_waza_p/common/patch.asm#L226).
Posting this in case I don't get to this soon or if someone else beats me to it.

theCapypara:
TIL the game has a debug crash screen in the retail version.

Adex-8x:
Oh, small correction and a bit more info: This crash happens when applied alongside `DungeonInterrupt`. The call that causes this seems to be around [here](<https://github.com/SkyTemple/skytemple-files/blob/master/skytemple_files/_resources/patches/asm_patches/anonymous_asm_mods/dungeon_interrupt/common/patch_arm9.asm#L55>)
Though looking at that code, I'm a bit confused by
```
	bl GetMovePPWithBonus
	strb r0,[r5,#+0x6]
	add r4,r4,#1
	add r5,r5,#8
	blt loop_move
```
Is it intended behavior for this patch to use the condition codes that `GetMovePPWithBonus` ends with...?
Trying to implement a stricter loop (comparing for 4 and checking the move's validity bit) either results in no visible effect or an earlier crash.

PR #464, opened 2023-12-31
Fix DungeonInterrupt loop

Adex-8x:
Adds a validity and a fixed bounds check to a loop in `DungeonInterrupt`, which caused crashes on hardware while applied alongside `ExternalizeWazaFile`.
Closes https://github.com/SkyTemple/skytemple-files/issues/463.


theCapypara:
Thanks! And happy new year!

PR #466, opened 2024-01-04
Add NegativeEnemyDensity patch

End45:
Continuation from https://github.com/SkyTemple/skytemple-files/pull/462#issuecomment-1872531275
Adds a new patch that makes negative enemy densities work as intended


theCapypara:
Thanks!

PR #468, opened 2024-01-07
Fix Orb oversight with Complete Team Control

Adex-8x:
When using Complete Team Control, if you attempted to use an Orb with an ally who wasn't originally the leader, the game would still recalculate the potential targets surrounding them. This has the noticeable impact of an ally using an Orb in an unintended direction from where they're currently facing. 
This should fix that behavior to make it so allies won't recalculate their targets.


theCapypara (approved):
Thanks!

PR #476, opened 2024-01-16
Swap effect and duration parameters for back_SetEffect and back2_SetEffect

Chesyon:
The effect is the first parameter, and the duration is the second parameter. They were the wrong way around (according to the wiki)


theCapypara:
Thanks!

PR #484, opened 2024-02-22
Wonder Mail S Generator

theCapypara:
Tests missing.


theCapypara:
will probably abandon this, somebody else started an implementation in Python and the bitstream stuff ended up shooting myself in the foot, should have just done bit masks.

PR #495, opened 2024-04-10
Consolidate string blocks across NA/EU/JP versions

in2erval:
This PR aims to revise the string blocks that are currently defined for the three Explorers of Sky versions (NA/EU/JP). The text linking was done via spreadsheet and added to the directory for future reference as `linked-eos-strs.csv`.

This new set of blocks have been compared with the previous version to ensure the regions being used by the UI (https://github.com/SkyTemple/skytemple/blob/e90441e56af1e0523254efe3498161b8981ac771/skytemple/core/string_provider.py#L32-L59) are named the same - there are some changes which I believe should be applied however:

### NA
* IQ Skills Names: **(9957, 10025) -> (9956, 10025)** 
  * ID 9957 should be included as the equivalent JP string is "[M:D1]かしこさ なし" meaning "No IQ"
* Trap Names: **(13504, 13530) -> (13504, 13529)** 
  * ID 13530 should not be included as it's a Trap Description

### EU
* IQ Skills Names: **(9959, 10027) -> (9958, 10027)** 
  * Same reason as NA

### JP
* IQ Skills Names: **(16147, 16216) -> (16146, 16215)** 
  * Same reason as NA, but also ID 16216 should not be included as it's an IQ Skills Description
* Move Descriptions: **(16286, 16844) -> (16286, 16845)** 
  * ID 16845 should be included as it's included in other versions - NA 10704, EU 10706
* Item Long Descriptions: **(16844, 18245) -> (16845, 18245)** 
  * ID 16845 should not be included as it's a Move Description

### DIFF
--- a/skytemple_files/_resources/ppmdu_config/pmd2data.xml
+++ b/skytemple_files/_resources/ppmdu_config/pmd2data.xml
@@ -87,86 +87,176 @@
         </Language>
       </Languages>
       <StringBlocks>
-        <StringBlock name="Explorer Ranks Names"                   beg="374"  end="387"/>
-        <StringBlock name="Default Team Names"                     beg="561"  end="567"/>
-        <StringBlock name="Job Debriefing Related Strings"         beg="606"  end="672"/>
-        <StringBlock name="Member Joined Menu"                     beg="676"  end="682"/>
-        <StringBlock name="Some Item Menu Strings"                 beg="682"  end="696"/>
-        <StringBlock name="Gummi Eating Strings"                   beg="696"  end="708"/>
-        <StringBlock name="Items and Pokemon Menu Strings"         beg="711"  end="777"/>
-        <StringBlock name="Kangaskhan Storage Strings"             beg="783"  end="812"/>
-        <StringBlock name="Evolution Strings"                      beg="1068"  end="1097"/>
-        <StringBlock name="Music tracks names"                     beg="1276"  end="1418"/>
-        <StringBlock name="Personality Quiz Questions"             beg="1418"  end="1485"/>
-        <StringBlock name="Personality Quiz Answers"               beg="1485"  end="1660"/>
-        <StringBlock name="Personality Quiz result strings"        beg="1660"  end="1708"/>
-        <StringBlock name="Personality string natures"             beg="1737"  end="1741"/>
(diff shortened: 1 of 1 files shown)


theCapypara:
Thanks! If we change existing strings then this is a breaking change but its probably worth it. Thanks for documenting it!

I'll review and test this today or tomorrow and then merge. Can you also open a new PR for the Randomizer for your smart text replacements to use the new string regions?

theCapypara:
@End45 Since this is quite the substential change I also added you as reviewer. Could you also have a look & check if any patch would break?

in2erval:
> I'll review and test this today or tomorrow and then merge. Can you also open a new PR for the Randomizer for your smart text replacements to use the new string regions?

Yep, I can get that PR going as well.

in2erval:
> I don't get this. String #9959 is the empty string before IQ skill names, which is already included and makes sense to include since, as you said, it represents "No IQ skill". But 9958 is "The Pokémon becomes afflicted with a\n[LS:13756]Terrified[LE] status. It will flee from\nenemies and friends, and it can't use\nitems while fleeing.\n\n\n\n\nSelect detail:[CLUM_SET:70][M:B4] or [M:B5]/View detail: [M:B2]\0". Are you sure those numbers are right?

The way that `beg` and `end` works in the XML seems to be that it selects strings based on `beg < string_id <= end` - 9958 is not actually part of IQ Skills Names:

image

This is likely due to the fact that string_id being indexed from 1 instead of 0 - let me know if this interpretation is incorrect.

theCapypara:
That is correct, the string IDs in SkyTemple are 1-indexed since the game also uses 1-indexing (effectively).

theCapypara:
> @End45 How would this break patches exactly?

I thought there were some patches that used strings, but maybe not.

End45:
> The way that beg and end works in the XML seems to be that it selects strings based on beg < string_id <= end

Huh. That doesn't make much sense (even if the strings are 1-indexed), but I see. The numbers are correct then.

> I thought there were some patches that used strings, but maybe not.

Some ASM patches might directly load strings by their ID, but those are not affected by changes to how SkyTemple groups them. So I think it should be alright.

theCapypara (approved):
LGTM! Thanks again!

End45 (approved):

PR #498, opened 2024-05-10
Fix oversight on DisplayScriptVariable

Adex-8x:
Fixes a small oversight with `DisplayScriptVariable` that might result in a crash on hardware; implements a proper parameter count check for the text tag.


Adex-8x:
No clue why the tests are failing btw, I tested myself and it applies just fine.

theCapypara:
Tests broke.

PR #503, opened 2024-05-22
Revert back_SetBackScrollSpeed param type

### LINKED ISSUE #504: back_SetBackScrollSpeed sint14, not sint16
Version: 1.6.5 Windows

SkyTemple ExplorerScript seems to be incorrectly assigning parameter 0 of the opcode "back_SetBackScrollSpeed()" as a 14 bit sint instead of a 16 bit sint, despite the .xml clearly indicating a sint16. This can be seen clearly in m24a0205, which seems to display both u16s and negative integers in it's script when opened on the latest version.

Pic unrelated
image

Adex-8x:
Reverts the opcode `back_SetBackScrollSpeed` to have parameter types of 16-bit signed ints (was accidentally changed when the wiki's info was applied here).

Closes #504

### DIFF
--- a/skytemple_files/_resources/ppmdu_config/pmd2scriptdata.xml
+++ b/skytemple_files/_resources/ppmdu_config/pmd2scriptdata.xml
@@ -1421,8 +1421,8 @@
           <Argument id="1" type="sint" name="offset1"/>
         </OpCode>
         <OpCode id="0x4"   name="back_SetBackScrollSpeed"                params="2"  stringidx="-1" unk2="0"  unk3="0"   >
-          <Argument id="0" type="sint" name="speed0"/>
-          <Argument id="1" type="sint" name="speed1"/>
+          <Argument id="0" type="sint16" name="speed0"/>
+          <Argument id="1" type="sint16" name="speed1"/>
         </OpCode>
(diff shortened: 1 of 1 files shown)


End45:
This is interesting. The parameters were originally annotated as `sint16` and `uint`. That info was then used to auto-generate the wiki page.
You were the one who changed it to `sint`, `sint`: https://wiki.skytemple.org/index.php?title=List_of_Opcodes&oldid=147, which then Para reflected on the XML: https://github.com/SkyTemple/skytemple-files/commit/529af17c167550757538364d3bff1d928bc659a4.
Now it's `sint16`, `sint16`? How is this being determined, exactly?
Technically, all those options are wrong, since the opcodes for this parameter are read using `ScriptParamToInt` (https://github.com/UsernameFodder/pmdsky-debug/blob/71542670fc1d905f406f910380a28b17081b9f97/symbols/overlay11.yml#L79), so their actual type is `fixed point flag (1 bit) + sign bit (1 bit) + 14 data bits`, but SkyTemple can't read that right now. Maybe `uint` would work better as a workaround, since it makes it easier to translate between what SkyTemple uses and the real underlying format.

theCapypara:
I opened a Discord thread for further discussion, the results will be documented here or at #504 .
https://discord.com/channels/710190644152369162/990722852200284180/1242961123527888988

theCapypara:
superseded by #508

PR #506, opened 2024-05-31
bump pmdsky-debug

theCapypara:
Some symbols were renamed, this still needs to be fixed, and other SkyTemple sources must be checked as well.

PR #507, opened 2024-05-31
Update Symbol instantiations

End45:
The format of the Symbol class has changed, so these instantiations need to be updated by adding the name and type of the symbols.


End45:
Marking as draft for now because this won't be required until #505 is merged and the pmdsky-debug-py version is bumped.

theCapypara:
Hi! I already did this in #506. You can add commits there if you want.

PR #523, opened 2024-07-22
Single hash table lookup for `map_back`

tech-ticks:
Replaces string operations in `map_back` with a pre-populated map of all constants.

### DIFF
--- a/skytemple_files/common/ppmdu_config/script_data.py
+++ b/skytemple_files/common/ppmdu_config/script_data.py
@@ -21,10 +21,11 @@ from __future__ import annotations
 #
 #  You should have received a copy of the GNU General Public License
 #  along with SkyTemple.  If not, see <https://www.gnu.org/licenses/>.
+from contextlib import contextmanager
 import re
 import warnings
 from enum import Enum, IntEnum
-from typing import Sequence, Mapping, Union
+from typing import Iterator, Sequence, Mapping, Union
 
 from explorerscript.ssb_converting.ssb_data_types import SsbCoroutine, SsbOpCode
 from range_typed_integers import i16, u8, u16
@@ -377,7 +378,7 @@ def camel_to_screaming_snake_case(string) -> str:
     return CAMEL_REGEX.sub("_", string).upper()
 
 
-class Pmd2ScriptData(AutoString):
+class _ScriptDataStorage(AutoString):
     def __init__(
         self,
         game_variables_table: list[Pmd2ScriptGameVar],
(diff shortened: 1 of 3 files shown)


skytemple_files/common/ppmdu_config/script_data.py
```diff
@@ -299,14 +300,7 @@ def __init__(self, ssa_id: int, name: str, ssb_id: int | None = None):
...
     def id(self) -> int:
-        """
-        For backwards compatibility.
-        """
-        warnings.warn(
-            DeprecatedToBeRemovedWarning("Please use self.ssa_id instead.", (2, 0, 0)),
-            stacklevel=2,
-        )
-        return self.ssa_id
+        return self.ssb_id
```
tech-ticks, line 303: I'm not sure if this change is safe, but it'd allow us to get rid of `SsbScriptDirection`, which saves a type check.

theCapypara: I don't really get this change, can't you just use `ssa_id` or `ssb_id`? If anything this should either be removed or kept because there is still code expecting this to return `ssa_id`, in which case everything would break.

theCapypara: oh you need it so it's compatible with SsbScriptDirection, Let me check the code...

theCapypara: Nah, we really shouldn't just change it like this, it's too risky... Maybe there's still ways to remove `SsbScriptDirection` but not use `Pmd2ScriptDirection.id` either

skytemple_files/common/ppmdu_config/script_data.py
```diff
@@ -364,6 +391,14 @@ def __init__(
         self.op_codes = op_codes
         self.ground_state_structs = ground_state_structs
 
+    def _insert_constants_with_prefix(self, vars: Sequence[ScriptDataConstant], prefix: str) -> None:
```
theCapypara, line 394: This is technically not entirely OK, since the setters would insert new constants, ie. for new actors or levels but not remove removed or renamed ones. Any idea what we could do about that?

theCapypara:
@tech-ticks Wouldn't it be easier to make it so that we just collect the dict of all possible constant values once during script compilation and then use that during the process? Instead of always globally caching the values?

tech-ticks:
Hm, that would also work. The other idea I had was to create the lookup table in the `ScriptData` constructor and add a function to rebuild it that has to be called after changing any data with a setter. Your suggestion would add a bit of overhead to each compiler invocation, but it might not matter much.

skytemple_files/common/ppmdu_config/script_data.py
tech-ticks: I found another way that should work.

tech-ticks:
Done. I think the actor list loader will have to call `rebuild_constant_lookup` now, and maybe you're aware of other places I might be missing @theCapypara?

tech-ticks:
Also, I'm sorta confused by these type errors:
```
skytemple_files/common/ppmdu_config/script_data.py:421:9: error: Incompatible types in assignment (expression has type "Pmd2ScriptFaceName", variable has type "Pmd2ScriptObject")  [assignment]
skytemple_files/common/ppmdu_config/script_data.py:435:13: error: Unsupported target for indexed assignment ("Mapping[str, Pmd2ScriptEntity | Pmd2ScriptObject | Pmd2ScriptRoutine | Pmd2ScriptFaceName | Pmd2ScriptFacePositionMode | Pmd2ScriptGameVar | Pmd2ScriptLevel | Pmd2ScriptMenu | Pmd2ScriptSpecial | Pmd2ScriptDirection | Pmd2ScriptDirectionSsb | Pmd2ScriptBgm | Pmd2ScriptSpriteEffect]")  [index]
skytemple_files/common/ppmdu_config/script_data.py:439:13: error: Unsupported target for indexed assignment ("Mapping[str, Pmd2ScriptEntity | Pmd2ScriptObject | Pmd2ScriptRoutine | Pmd2ScriptFaceName | Pmd2ScriptFacePositionMode | Pmd2ScriptGameVar | Pmd2ScriptLevel | Pmd2ScriptMenu | Pmd2ScriptSpecial | Pmd2ScriptDirection | Pmd2ScriptDirectionSsb | Pmd2ScriptBgm | Pmd2ScriptSpriteEffect]")  [index]
```

theCapypara:
> Done. I think the actor list loader will have to call `rebuild_constant_lookup` now, and maybe you're aware of other places I might be missing @theCapypara?

Whenever the actor and level list are updated in SkyTemple



> Also, I'm sorta confused by these type errors:
> 
> ```
> skytemple_files/common/ppmdu_config/script_data.py:421:9: error: Incompatible types in assignment (expression has type "Pmd2ScriptFaceName", variable has type "Pmd2ScriptObject")  [assignment]
> skytemple_files/common/ppmdu_config/script_data.py:435:13: error: Unsupported target for indexed assignment ("Mapping[str, Pmd2ScriptEntity | Pmd2ScriptObject | Pmd2ScriptRoutine | Pmd2ScriptFaceName | Pmd2ScriptFacePositionMode | Pmd2ScriptGameVar | Pmd2ScriptLevel | Pmd2ScriptMenu | Pmd2ScriptSpecial | Pmd2ScriptDirection | Pmd2ScriptDirectionSsb | Pmd2ScriptBgm | Pmd2ScriptSpriteEffect]")  [index]
> skytemple_files/common/ppmdu_config/script_data.py:439:13: error: Unsupported target for indexed assignment ("Mapping[str, Pmd2ScriptEntity | Pmd2ScriptObject | Pmd2ScriptRoutine | Pmd2ScriptFaceName | Pmd2ScriptFacePositionMode | Pmd2ScriptGameVar | Pmd2ScriptLevel | Pmd2ScriptMenu | Pmd2ScriptSpecial | Pmd2ScriptDirection | Pmd2ScriptDirectionSsb | Pmd2ScriptBgm | Pmd2ScriptSpriteEffect]")  [index]
> ```

1. You are re-using the name `var`, `mypy` doesn't like that. just use another variable name for subsequent loops
2. and 3. `Mapping` is not mutable. You need to declare `_all_script_constants__by_name` as `MutableMapping`. Or just as a `dict`. `Mapping` and `MutableMapping` really just make sense for public types to restrict the API contract.

theCapypara:
Also @tech-ticks in `loader.py` of skytemple-files. After the data has been loaded from the ROM.
You can look for all usages of the setter with an IDE, just make sure that SkyTemple and the Randomizer are included in the usage.

theCapypara:
It seems that rebuilding the constants after loading ROM data doesn't work. The ROM tests fail now complaining that all face constants seem to be unknown. (Only the `false` ROM tests test SSB (de-)compilation, so that's why only they are missing).

theCapypara (approved):

PR #526, opened 2024-07-27
File API V2: waza_p asset

AnonymousRandomPerson:
Based off https://github.com/SkyTemple/skytemple-files/pull/511, but rebased off master because the original branch has conflicts.

Most of the stubs have been filled in, aside from:
- Custom Skypatches
- Asset hash matching (not sure what the use case is for these, see below)
- Symbol/Patch info assets (not sure what these are intended to be)
- CLI
- Documentation
I'm putting out this draft PR to see if I'm on the same page as others before I work on the CLI/documentation.

I implemented serialization/deserialization of `waza_p.bin` (moves/learnsets) as a proof of concept, along with a test script (`extract_rom_files_to_project.py`) that creates an asset project by extracting the waza_p from the ROM. The asset files are in JSON and almost a 1-1 match with the waza_p structs within Python, other than enums which are serialized to their string names if possible.

I don't really understand the use case for asset hash matching. My understanding is that this is intended as a checksum to prevent external modification of the assets outside the SkyTemple ecosystem. But one of the advantages of being human-readable is that this allows people to easily modify the files without needing a specialized program like SkyTemple. Having a checksum would nullify this advantage.

### DIFF
--- a/skytemple_files/common/file_api_v2.py
+++ b/skytemple_files/common/file_api_v2.py
@@ -123,9 +123,11 @@ class RomProject:
     @property
     def patcher(self):
         """
-        Returns a patcher. May load aribitrary code from SkyPatches, see `self.does_allow_extra_skypatches`.
+        Returns a patcher. May load arbitrary code from SkyPatches, see `self.does_allow_extra_skypatches`.
         This may be cached until `self.set_allow_extra_skypatches` was called.
         """
+        if self.rom is None:
+            raise ValueError("No ROM available to patch")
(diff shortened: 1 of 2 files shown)


theCapypara:
@AnonymousRandomPerson The hash serves two purposes:
- Detecting when a file in the ROM was changed without a change in the ROM
- Detecting when an asset was changed without changes in the ROM

Both are done to control the import process and to provide the app/CLI/user the possibility to decide how to deal with the situation.

PR #527, opened 2024-07-27
Move Sprite Size elements to another patch

irdkwia:
Sprite Size part of the ExpandPokeList patch is independent an can be extracted to another patch.
This changes ExpandPokeList so that it depends on that new patch, SpriteSizeInMonsterData.


theCapypara:
Thanks!

Issue #248, opened 2022-06-14
Improve the Image chopping algorithm for sprite import.


audinowho:
https://github.com/SkyTemple/skytemple-files/blob/fae264783a10969678d6e2cebaaa46e4309549ef/skytemple_files/graphics/chara_wan/writer.py#L1357

When a Pokemon sprite wan file is imported, every frame gets chopped up into pieces such that it can be assembled by metaframe data.
The method responsible for this is called chopImgToPieceLocs.
Pieces can be 8x8, 16x8, 8x16, etc.

Currently, the function just returns one piece containing the whole sprite if it's small enough (it usually is), or a collection of pieces that divide the sprite into as few pieces as possible.

This approach does not take advantage of the ability for one piece to be referenced in multiple parts of a sprite.  For example: the symmetrical legs of a large Pokemon can be chopped into their own pieces and then collapsed into a single leg texture that is drawn normally for the left side, and then flipped for the right side.  This can spell the difference between a sprite being too big to import, or being just small enough to import.

![image](https://user-images.githubusercontent.com/2676737/173675799-307a169a-bf05-4a74-86e1-1af69c54464f.png)

audinowho:
A potential (naive) method would be as follows:

Iterate through every possible position by which you can put an 8x8 tile on the image, including partial outside positions.
This means (imgWidth + (8 - 1) * 2) * (imgHeight + (8 - 1) * 2) positions.

For each possible position, iterate through the entire thing a second time to see if any other positions have pixels that match the first position.
This means another (imgWidth + (8 - 1) * 2) * (imgHeight + (8 - 1) * 2) positions.

And you will need to compare flipW, flipH, and flipW+H versions of the image too.
So multiply it by 4.

Once all matching images for a given first position have been gathered, store them in a variable collection.
It is possible to optimize out repeat first-images by keeping them in a dictionary that maps position to position-copies.

It may also be possible to apply this to all frames instead of just this frame, but doing so will require refactoring outside of this function call.  It will greatly increase optimization but also increase time by a factor of n^2 where n is the number of images.

After this mass iteration, you will have a list of image similarity data.
Pick the combination of image groups that compresses the most pixels, in which overlaps cannot occur.  Greedy algorithm or whatever else you can think of.
You can opt to remove overlaps by marking them when choosing images.

A frame collection of 64x64 pixels will require this many comparisons:
[...]

PR #530, opened 2024-07-28
Improve Sprite Import Algorithm

irdkwia:
This improves frame chunking algorithm to use minimal amount of blocks in VRAM for minimal Sprite Size.
This algorithm has some effects on the Sprite File Size, tends to slightly increase it but can also slightly decrease it.

marius851000/pmd_wan#4 should also be merged to fix wrong maximum amount of blocks, which tends to overestimate Sprite Size value.

See #248

Contains also changes to Sprite Size calculation, so this should be merged at the same time as Skytemple/skytemple#785


skytemple_files/graphics/chara_wan/sheets.py
```diff
@@ -831,8 +829,8 @@ def mapDuplicateImportImgs(imgs, final_imgs, img_map):
                     flip = final_idx
             else:
                 imgs_flip = exUtils.imgsEqual(final_img[0], img[0], True)
-                if imgs_flip:
-                    imgs_flip = exUtils.offsetsEqual(final_img[1], img[1], img[0].size[0], True)
+                # if imgs_flip:
```
audinowho, line 832: Can this be verified against multi-offset Pokemon like Jirachi?
Multi-offset Pokemon have multiple frames with the same textures but different offsets.  Will they still import correctly with this change?

irdkwia: This should work with differing offsets, because the algorithm already have this system used for frames that are identical in texture but differing in offsets in the branch above that, I don't see why it should be restricted here.
As for Jirachi, there is indeed a problem with one of their offsets on Special2, but that problem is also present in the current version of the algorithm and may be linked with the animation frame for that offset being an empty frame.

PR #539, opened 2024-08-08
Sprite fixes and improvements

irdkwia:
- Fix sprite with missing chunks in frame
  - This is due to the Y offset being misinterpreted in chunks (Y offset is actually a signed 10-bit integer)
- Slightly improve sprite file size reduction
  - Reuses the same algorithm that merges chunks with same image data, which can slightly improve file size especially if two frames are similar in the sprite


theCapypara:
This fixes the issue of the bottom row being cut off right?

irdkwia:
> This fixes the issue of the bottom row being cut off right?

Yes

PR #543, opened 2024-08-18
Fix ExpandPokeList patch Unown strings

irdkwia:
Fixes a problem with Unown species using a hardcoded check with its own String ID


theCapypara (approved):
Thanks

PR #544, opened 2024-08-20
Fix Dungeon Interrupt HP bug

ShiningDeneb:
Since I worked on it, I wanted to submit a fix to the (old) bug of DungeonInterrupt patch where the entity HP is not synchronized with the tabulated new max HP value.

### DIFF
--- a/skytemple_files/_resources/patches/asm_patches/anonymous_asm_mods/dungeon_interrupt/common/patch_overlay11.asm
+++ b/skytemple_files/_resources/patches/asm_patches/anonymous_asm_mods/dungeon_interrupt/common/patch_overlay11.asm
@@ -25,7 +25,7 @@
 	mov  r1,r7
 	str r3,[r13]
 	str r9,[r13, #+0x8]
-	ldrsh r9,[r8, #+0xe]
+	nop
 	nop
 	ldr r3,[r15, #+0x80]
 	str r9,[r13, #+0x4]


Adex-8x:
Unfortunately, I don’t think this is a fix. The reason this code obtains a team member’s current HP is that HP isn’t automatically healed when exiting a dungeon via dungeon interruption. With your new `nop`, team members’ HP will display incorrectly if they exit a dungeon via dungeon interruption with less than their maximum HP.
I’m assuming this was done with the intent to try and fix Brick-Tough having a bit of a mismatch when displaying in the overworld, so while this technically fixes that bug, it also creates a bigger one.

irdkwia:
I don't approve these changes either, for one I side with what Adex already said, and because that doesn't fix other HP display (i.e. the summary in your team stats menu).
Also, the proposed changes are exactly the behaviour of the game without this patch, which doesn't care about HP in overworld because base game never need to save your current party's HP between dungeons.
That changes though with DungeonInterrupt since theoretically it introduces cases of no healing between dungeons/interruptions, in which the base game logic is not the one wanted.
So, either the logic behind Brick-Tough is changed, or the logic of the patch must be changed and drop the no healing between interruptions part, in that case the base game HP logic can be kept (though that may be implemented via parameters to let people choose between no healing/healing logics).

ShiningDeneb:
Okay I think I see the bigger issue you mention. I guess changing how Brick-Tough is accounted in the overworld sounds more preferable compared to changing the way the patch works
I will test that in the next days but I want your input on that though: do you think having Brick-Tough increase the entity current HP if it's equal to max HP does the trick?

Adex-8x:
I haven’t tried it myself, but I guess that could be an okay compromise? As well as accounting for anything else that could boost HP in the overworld, I’m not sure if Brick-Tough is the only thing.

PR #547, opened 2024-10-06
Add description translatability for some patches

Adex-8x:
Oops, I forgot to make my patches' descriptions translatable, so I tweaked them (plus one of Frostbyte's).
Is this enough, or is there more that's needed for the translation stuff to just work? I'm pretty unfamiliar with it.


theCapypara (approved):
Thanks!

PR #548, opened 2024-10-27
sma support

audinowho:
Parses /SYSTEM/manpu_su.sma, which contains status effect icon animations.
Only reads and exports.  Import not supported.


theCapypara (approved):
Thanks!

PR #549, opened 2024-10-27
effect_wan support

audinowho:
Parses wan files in /EFFECT/effect.bin, with some exceptions (see the README) which contains move VFX and cutscene VFX.
Only reads and exports. Import not supported.


theCapypara (approved):
Thanks!

PR #551, opened 2024-10-31
Screen effect support

audinowho:
Parses screen effect (which are not wan) files in /EFFECT/effect.bin, which contains move VFX and cutscene VFX.
These are specifically things like blizzard, hail, fog, cosmic power, heat wave graphics, etc.
Only reads and exports. Import not supported.


theCapypara (approved):
Thanks!

PR #556, opened 2024-12-14
FixNo$GbaSaves 1.0.0

Chesyon:
FixNo$GbaSaves no longer breaks saving on melonDS.
Related: https://github.com/Frostbyte0x70/EoS-asm-patches/pull/10

### DIFF
--- a/skytemple_files/common/exceptions/outdated_patch_dependency.py
+++ b/skytemple_files/common/exceptions/outdated_patch_dependency.py
@@ -1,0 +1,27 @@
+from skytemple_files.common.i18n_util import _
+
+
+class OutdatedPatchDependencyError(Exception):
+    """
+    Raised when trying to apply an ASM patch that depends on another patches that exist but are outdated
+    """
+
+    def __init__(self, patch_name: str, outdated_dependencies: list[str]):
+        """
+        Creates a new OutdatedPatchDependencyError
+        :param patch_name: Name of the patch that failed to apply due to outdated dependencies
+        :param outdated_dependencies: List containing the names of each outdated dependency
+        """
+        self.patch_name = patch_name
+        self.outdated_dependencies = outdated_dependencies
+
+    def __str__(self):
+        msg = _(
+            f"The patch {self.patch_name} cannot be applied because the following patches it depends on "
+            f"are outdated:"
+        )
+        for dependency in self.outdated_dependencies:
+            msg += f"\n- {dependency}"
+        msg += _("Please reapply these patches first.")
+
+        return msg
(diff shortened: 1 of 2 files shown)


skytemple_files/patch/handler/fix_nocash_saves.py
```diff
@@ -30,37 +30,38 @@
...
 
     @property
     def author(self) -> str:
-        return "Frostbyte"
+        return "Frostbyte, Chesyon"
 
     @property
     def version(self) -> str:
-        return "0.1.0"
+        return "0.2.0"
```
Frostbyte0x70, line 56: You know, since this patch now fully works with no side effects, why don't we bump it to v1.0 instead?

Frostbyte0x70 (commented):
Looks good!
The PR on my repo needs to be merged first, then the submodule needs to be updated. I'll approve once those two things are done.

Chesyon:
Some kind of check needs to be added for the ExtraSpace version; this patch crashes the game on boot on old ExtraSpace versions, because ExtraSpace isn't loaded early enough for the first hook in those versions.

Frostbyte0x70:
> Some kind of check needs to be added for the ExtraSpace version; this patch crashes the game on boot on old ExtraSpace versions, because ExtraSpace isn't loaded early enough for the first hook in those versions.

Hm... you're right. SkyTemple does not allow setting a minimum version of another patch as a dependency though, so not sure how this could be added.

Chesyon:
> Hm... you're right. SkyTemple does not allow setting a minimum version of another patch as a dependency though, so not sure how this could be added.
>
It would be kind of janky, but we could perhaps run a secondary is_applied check for the latest ExtraSpace version in patch.py, and prompt the user to reapply ExtraSpace if the check fails. I don't really know how to add prompts to the patch.py, though, so I might need some help in this regard.

theCapypara:
@Chesyon Some checks are failing.

theCapypara (changes requested):
- Check for ExtraSpace version needs to be implemented
- Checks are failing

Chesyon:
> * Check for ExtraSpace version needs to be implemented
> 
>     * Checks are failing
> 
>     * The submodule needs to be updated

My code isn't finished; Frostbyte volunteered to implement the prompt. Once the prompt is added, the checks _should_ pass;  it's only currently failing because there's nothing inside the if statement, causing Python's complaining about indents. The ExtraSpace version check is implemented; all we're waiting on is the prompt from Frostbyte.

Frostbyte0x70:
Huh? Why did that get closed? Did I screw something up when pushing?

theCapypara:
You force-pushed Chesyon:master to be the same as skytemple-files:master

Frostbyte0x70:
I was trying to push my local branch to chesyon:master
I guess Git didn't like that they had different names and pushed master instead
Oh well, new PR here: https://github.com/SkyTemple/skytemple-files/pull/558

PR #562, opened 2024-12-24
Chara wan fix

audinowho:
Fixes default animation labeling for grovyle.
Adds the option to delete anims from a wan file.


theCapypara:
Thanks!

PR #568, opened 2025-01-08
Add "Ai Item Improvements" and "Ai Item Optimizations" Skypatches to SkyTemple

HappyLappy1:
Ai Item Optimizations reduces the size of GetAiUseItemProbability without changing anything else, creating 1080 bytes of free space in overlay29.

Ai Item Improvements replaces the same function with one that tweaks team and enemy AI to use more types of items, and use them more effectively. Here's a full list of all the things it adds/modifies:
   - Made Gold Thorns and Rare Fossils Throwable by the Ai!
   - Fixed Gravelyrock bug that only checked male Bonsly and Sudowoodo.
   - Fixed Violent Seed Bug that only checked SPA instead of both ATK and SPA
   - Vile Seed now runs a check similar to Violent Seed to ensure foe is not at minimum defenses already
   - Via Seeds now behave like Warp Seeds
   - Gravelyrocks will be thrown unless the user or target is Sudowoodo/Bonsly
   - Favored Gummis (Resisted or Better) will be eaten. Wonder/Wander/Nectar will always be eaten!
   - Violent Seeds, Rebound Orbs, and All-Hit Orbs will be used at enemies over 3/4 HP!
   - Warp/Via Seeds, Gone Pebbles, Vanish Seeds, Invisify Orbs, Rollcall Orbs (Enemies only), and Pure Seeds will be used if below 1/4 HP!
   - Sleep Seeds now check for sleeplessness and yawning!
   - Chesto Berries will only ever be used if sleeping, nightmared, or yawning!
   - For Sticks and Stones, Team members have a throw chance that depends on the foe's HP!
       - Allies have a 5% chance, unless the foe has low HP. If they do, it's an 80% chance! 
       - Foes still have 70% throw chance.
   - Allies will always use Rollcall Orbs if given one! Enemies will use them at low health!
   - Grimy Food will be eaten at sufficiently low hunger, or thrown: 80% if the target is healthy, and 30% otherwise.
   - Will use Weather Orbs if it would change the weather, and they "benefit" from the weather change
       - Hail: Ice only | Sandstorm: Rock, Ground, Steel | Sunny: Fire, Grass  |Rainy: Water, Electric
   - Will use Keys if they are in front of a Key Door
   - Will use Drought Orbs if they cannot swim and secondary terrain is water!
   - Will eat Blast Seeds if next to enemies!
   - Will use Ginseng on themselves, rather than being forced to throw at an ally
   - Oren Berries can be used or thrown on any low HP pokemon! (The Ai may eat or throw them!!!)
   - Foes will only use Trawl Orbs if a Kecleon shop is NOT on the floor!

### DIFF
--- a/skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
+++ b/skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
@@ -30,7 +30,6 @@
 ;   - Will use Ginseng on themselves, rather than being forced to throw at an ally
 ;   - Oren Berries can be used or thrown on any low HP pokemon! (The Ai may eat or throw them!!!)
 ;   - Foes will only use Trawl Orbs if a Kec shop is NOT on the floor!
-;   - Secret Easter Egg if Species is Lapras :P
 
     stmdb sp!,{r3,r4,r5,r6,r7,r8,r9,lr}; Push a bunch of registers, we need room to "function"
     mov r7,r0; Keep our precious monster pointer safe
@@ -660,7 +659,7 @@ label1:
     cmpne r1,r2;
     addne r2,#8;
     cmpne r1,r2;
-    beq odds_100;
+    beq odds_0; Fine I'll remove it Ches
     b ItemChecks;
 
 IsHailOrb:
--- a/skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
+++ b/skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
@@ -1,0 +1,805 @@
+.org AiOrbCheck
+.area 0x4
+    cmp r9,#16; This will always return false. Used to be a check for whether the item-category was "Orbs", but is completely neutered! 
+.endarea
+
+.org GetAiUseItemProbability
+.area IsAdjacentToEnemy - GetAiUseItemProbability 
+; Changelog
+;   - Made Gold Thorns and Rare Fossils Throwable by the Ai!
+;   - Fixed Gravelyrock bug that only checked male Bonsly and Sudowoodo.
+;   - Fixed Violent Seed Bug that only checked SPA instead of both ATK and SPA
+;   - Vile Seed now runs a check similar to Violent Seed to ensure foe is not at minimum defenses already
+;   - Via Seeds now behave like Warp Seeds
+;   - Gravelyrocks will be thrown unless the user or target is Sudowoodo/Bonsly
+;   - Favored Gummis (Resisted or Better) will be eaten. Wonder/Wander/Nectar will always be eaten!
+;   - Violent Seeds, Rebound Orbs, and All-Hit Orbs will be used at enemies over 3/4 HP!
+;   - Warp/Via Seeds, Gone Pebbles, Vanish Seeds, Invisify Orbs, Rollcall Orbs (Enemies only), and Pure Seeds will be used if below 1/4 HP!
+;   - Sleep Seeds now check for sleeplessness and yawning!
+;   - Chesto Berries will only ever be used if sleeping, nightmared, or yawning!
+;   - For Sticks and Stones, Team members have a throw chance that depends on the foe's HP!
+;       - Allies have a 5% chance, unless the foe has low HP. If they do, it's an 80% chance! 
+;       - Foes still have 70% throw chance.
+;   - Allies will always use Rollcall Orbs if given one! Enemies will use them at low health!
+;   - Grimy Food will be eaten at sufficiently low hunger, or thrown: 80% if the target is healthy, and 30% otherwise.
+;   - Will use Weather Orbs if it would change the weather, and they "benefit" from the weather change
+;       - Hail: Ice types only   Sandstorm: Rock, Ground, Steel     Sunny: Fire, Grass      Rainy: Water, Electric
+;   - Will use Keys if they are in front of a Key Door
+;   - Will use Drought Orbs if they cannot swim and secondary terrain is water!
+;   - Will eat Blast Seeds if next to enemies!
+;   - Will use Ginseng on themselves, rather than being forced to throw at an ally
+;   - Oren Berries can be used or thrown on any low HP pokemon! (The Ai may eat or throw them!!!)
+;   - Foes will only use Trawl Orbs if a Kec shop is NOT on the floor!
+;   - Secret Easter Egg if Species is Lapras :P
+
+    stmdb sp!,{r3,r4,r5,r6,r7,r8,r9,lr}; Push a bunch of registers, we need room to "function"
+    mov r7,r0; Keep our precious monster pointer safe
+    and r5,r2,0x1; If r5 is 0x1, the item is being thrown!
+    and r4,r2,0x2; If r4 is 0x2, the item is being used on an enemy!
+    ldr r6,[r7,#0xb4]; Extract the monster pointer, or smthn. We'll use r6 to check the monster's data  
+    mov r3, r1; Retain the Item Pointer in case we need it!
+    ldrsh r1,[r1,#0x4]; Extract the Item ID.
+    b label1
+ItemChecks:
+    cmp r1,#0x0a; Sticks and Stones
+        ble SticksAndStones; If Throwing, 70% (If teammate, 20% + 0.5 * ItemsHeld)
+
+    cmp r1,#0x0e; Y-Ray Specs
+    cmpne r1,#0x0f; Gaggle Specs
+        beq IsGaggleOrYRaySpecs; If Throwing, 80%
+
+    cmp r1,#0x14; Patsy Band
+        beq odds_40; In general, 40% (shouldn't be usable though)
+
+    cmp r1,#0x22; Diet Ribbon
+        beq IsDietRibbon; If Throwing at non-zero belly, 50%
+
+    cmp r1,#0x2F; Whiff Specs
+    cmpne r1, 0x30; No-Aim Scope
+        beq IsWhifforNoAim; If Throwing, 50%
+
+    cmp r1,#0x45; Heal Seed
+    cmpne r1,0x5C; Gabite Scale
+        beq IsHealSeed; If Negative Status, 80%
+
+    cmp r1,#0x46; Oran Berry
+    cmpne r1,#0x47; Sitrus Berry
+    cmpne r1,#0x75; Oren Berry
+        beq IsOranOrSitrusBerry; If HP >= Max, 0%. If Below Quarter health...
+        ; If below 1/4 and Throwing: 50%
+	    ; If below 1/4 and Using: if adjacent to Enemy 100%, else 50%
+
+    cmp r1,#0x48; EyeDrop Seed
+        beq IsEyeDropSeed; If not EyeDropped (Goggle Specs counts!): If adjacent to enemy, 80%, else 5%.
+    
+    cmp r1,#0x4a; Blinker Seed
+        beq IsBlinkerSeed; If not blinded: If adjacent to enemy, 80%, else 5%.
+    
+    cmp r1,#0x4B; Doom Seed
+        beq IsDoomSeed; If not level 1, 80%
+
+    cmp r1,#0x4c; X-Eye Seed
+        beq IsXEyeSeed; If not Cross-Eyed: If adjacent to enemy, 80%, else 5%.
+    
+    cmp r1,#0x4d; Life Seed
+        beq IsLifeSeed; If Thrown: If Adjacent to Enemy, 10%, else 100%.
+    
+    cmp r1,#0x4e; Rawst Berry
+        beq IsRawstBerry; If Burned, 50% chance.     
+    
+    cmp r1,#0x4f; Hunger Seed
+        beq IsHungerSeed; If non-zero belly, 50% chance.
+    
+    cmp r1,#0x50; Quick Seed
+        beq IsQuickSeed; If +3 or slower: if Adjacent to enemy, 80%, else 5%.   
+    
+    cmp r1,#0x51; Pecha Berry
+        beq IsPechaBerry; If Poisoned or Badly Poisoned: If Adjacent to enemy, 50%, else 100%.
+    
+    cmp r1,#0x52; Cheri Berry
+        beq IsCheriBerry; If Paralyzed: If Adjacent to enemy, 30%, else 80%.
+    
+    cmp r1,#0x53; Totter Seed
+        beq IsTotterSeed; If Not Confused: If Adjacent to enemy, 15%, else 80%.
+    
+    cmp r1,#0x54; Sleep Seed
+        beq IsSleepSeed; If Not immune to sleep or sleeping (Sleepless, Nightmared, Napping, Yawning, Asleep), Adjacent 80/5 Use!
+
+    cmp r1,#0x56; Warp Seed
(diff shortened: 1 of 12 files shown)


assidion2:
Maybe you can specify where the free space begins in the description?

skytemple_files/_resources/ppmdu_config/pmd2data.xml
```diff
@@ -1377,6 +1377,23 @@
...
+	  <OpenBin filepath="overlay/overlay_0029.bin">
+            <Include filename ="happylappy_asm_mods/ai_item_optimizations/selector_ov29.asm"/>
+          </OpenBin>
+        </Patch>
+        <Patch id="AiItemImprovements">
+	  <OpenBin filepath="overlay/overlay_0029.bin">
+            <Include filename ="happylappy_asm_mods/ai_item_improvements/selector_ov29.asm"/>
+          </OpenBin>
+	  <Parameters>
+	     <Param name="NeutralGummiBoost" label="Amount of IQ a neutral Gummi gives (If larger, Ai will eat the gummi!)" min="0" max="255" default="3" type="int"/>
```
Frostbyte0x70, line 1390: Can't you check this with ASM? Seems annoying for the user to check it manually, and to update it manually if the value changes (which requires reapplying the patch)
Also, what if neutral gummies increase the IQ by a different amount for each type? I believe it's possible to do, since the amounts are stored in a table.

skytemple_files/_resources/ppmdu_config/pmd2data.xml
```diff
@@ -1377,6 +1377,23 @@
...
+            <Include filename ="happylappy_asm_mods/ai_item_optimizations/selector_ov29.asm"/>
+          </OpenBin>
+        </Patch>
+        <Patch id="AiItemImprovements">
+	  <OpenBin filepath="overlay/overlay_0029.bin">
+            <Include filename ="happylappy_asm_mods/ai_item_improvements/selector_ov29.asm"/>
+          </OpenBin>
+	  <Parameters>
+	     <Param name="NeutralGummiBoost" label="Amount of IQ a neutral Gummi gives (If larger, Ai will eat the gummi!)" min="0" max="255" default="3" type="int"/>	
+            <Param name="AssignAiFlags" label="Assign AI flags (idk explanify further)" type="select">
```
Frostbyte0x70, line 1391: Should have a better description

skytemple_files/patch/handler/ai_item_improvements.py
```diff
@@ -0,0 +1,204 @@
...
+    @property
+    def author(self) -> str:
+        return "happylappy"
+
+    @property
+    def version(self) -> str:
+        return "1.0.0"
+
+    def depends_on(self) -> list[str]:
+        return []
```
Frostbyte0x70, line 71: Doesn't the patch depend on AiItemOptimizations?

skytemple_files/patch/handler/ai_item_optimizations.py
```diff
@@ -0,0 +1,83 @@
...
+from skytemple_files.common.i18n_util import f, _
+
+MODIFIED_INSTRUCTION = 0xE8BD83F8  # ldmia all the shit
+OFFSET_EU = (
+    0x231F358 - 1084 - 0x22DCB80
+)  # location of the next function minus the number of saved bytes minus 4 minus starting point of overlay 29
+OFFSET_US = 0x231E8F0 - 1084 - 0x22DC240  # same but US HeHeHaHa
+
+
+class AiItemOptimizationsHandler(AbstractPatchHandler, DependantPatch):
```
Frostbyte0x70, line 43: Shouldn't be a DependantPatch if it doesn't depend on anything

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_optimizations/common/optimizations_ov29.asm
```diff
@@ -0,0 +1,402 @@
...
+odds_50:
+    add r0,#10;
+odds_40:
+    add r0,#10;
+odds_30:
+    add r0,#20;
+odds_10:
+    add r0,#5;
+odds_5:
+    add r0,#5;
```
Frostbyte0x70, line 397: This is quite confusing to follow. Why not replace `b odds_xxx` with `mov r0,xxx; b exit`? Do you need the extra few instructions that badly?
If so, I'd at least add a comment at the start of this section explaining what it does.

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_optimizations/common/optimizations_ov29.asm
```diff
@@ -0,0 +1,402 @@
...
+    beq odds_50;    
+    b odds_0;
+
+; These lines are stupidly common. Might as well make a proper function!
+Adjacent2Enemy:
+    push r14;
+    mov r0,r7; 
+    bl IsAdjacentToEnemy
+    cmp r0,#0x1;
+    mov r0,#0;
```
Frostbyte0x70, line 158: Why the unconditional mov? I see you're effectively returning the result of the comparison in the status register (cursed), but in that case, you don't need to set r0, right?

Edit: Oh, it's because you might jump to something like odds_50. You might want to add a comment here, it's not very obvious.

skytemple_files/patch/handler/ai_item_improvements.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def category(self) -> PatchCategory:
+        return PatchCategory.NEW_MECHANIC
```
Frostbyte0x70, line 75: "The improvement/tweak" category would probably work better

skytemple_files/patch/handler/ai_item_improvements.py
```diff
@@ -0,0 +1,204 @@
...
+                )
+            else:
+                raise NotImplementedError()
+        else:
+            raise NotImplementedError()
+        if add_types_applied:
+            param["AddTypesApplied"] = 1
+        else:
+            param["AddTypesApplied"] = 0
+        self.__parameters = param
```
Frostbyte0x70, line 106: Since you're normally not supposed to overwrite this, I'd add a comment explaining why your're doing it

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
```diff
@@ -0,0 +1,804 @@
...
+        subgt r1,#0x2; Put Fairy Gummi in the right slot
+        beq odds_80; If Wonder Gummi, 80% use chance!
+        sub r1,#0x76; Convert Item ID to Gummi Type
+        ;lsl r1,#0x1; Double Gummi Type to increment by half-words
+        ldrb r2,[r6,#0x5E]; Get type 1
+        ldrb r3,[r6,#0x5F]; Get type 2
+        ldr r4,=IQ_GUMMI_GAIN_TABLE;
+        mov r5,#0x19; Waste a register and an instruction bc MUL doesn't allow immediates :(
+        ; Gummi Type 1
+        smulbb r2,r2,r5; Multiply by 18 bytes (?)
```
Frostbyte0x70, line 765: The immediate above says 19

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
```diff
@@ -0,0 +1,804 @@
...
+        cmp r1,#0x88; Wonder Gummi
+        beq odds_80; If Wonder Gummi, 80% use chance!
+        sub r1,#0x76; Convert Item ID to Gummi Type
+        lsl r1,#0x1; Double Gummi Type to increment by half-words
+        ldrb r2,[r6,#0x5E]; Get type 1
+        ldrb r3,[r6,#0x5F]; Get type 2
+        ldr r4,=IQ_GUMMI_GAIN_TABLE;
+        mov r5,#0x24; Waste a register and an instruction bc MUL doesn't allow immediates :(
+        ; Gummi Type 1
+        smulbb r2,r2,r5; Multiply by 18 bytes (?)
```
Frostbyte0x70, line 785: The immediate above says 24

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
```diff
@@ -0,0 +1,804 @@
...
+        cmp r2, NeutralGummiBoost;
+    .endif
+    bgt odds_80; 80% Chance
+    b odds_0;
+
+.pool; Constants will go here. Currently 4 things in the pool!
+; Put at the end because it has a variable amount of instructions.
+
+FreeSpaceStart:
+    .fill IsAdjacentToEnemy - FreeSpaceStart, 0
```
Frostbyte0x70, line 803: I'd consider directly using address of `IsAdjacentToEnemy` here, since you're not really referring to the function, just to the exact offset where your area ends, which happens to match the start of `IsAdjacentToEnemy`.
What if the offset of that function is changed, or it's renamed? It would be confusing, at least.

Frostbyte0x70 (changes requested):
I have a couple of gripes with the idea of adding these patches by default tbh

- First, there's no way to ensure that AiItemOptimizations doesn't break anything. It's basically impossible to test that all previous AI behavior always works the same way, and if there's a bug, it's going to be basically impossible to spot.
- Second, AiItemImprovements seems like a very specific patch that I don't see many people wanting to apply. I think it makes more sense as a community skypatch.
- Third, you can't properly list everything AiItemImprovements does in the description, since it's too long. People won't want to apply a patch without knowing what it does, which requires checking out the patch on the community patches channel anyway, so why not keep it there?

I still pointed out a few things you need to fix, just in case.

(Note to self / other reviewers: I haven't tested if the patch applies correctly yet)

skytemple_files/patch/handler/ai_item_improvements.py
HappyLappy1: It actually doesn't. Both patches write over the entirety of the original function, and can be applied on top of each-other without issue.

skytemple_files/_resources/ppmdu_config/pmd2data.xml
HappyLappy1: That's true. I could rename this to "gummi eat threshold" or something I suppose? I could read this value with ASM, but as you say, if I just read one neutral gummi interaction, we run into problems where it might not be the same for the others.

skytemple_files/patch/handler/ai_item_optimizations.py
HappyLappy1: So I just remove that field? Will do!

HappyLappy1:
> * First, there's no way to ensure that AiItemOptimizations doesn't break anything. It's basically impossible to test that all previous AI behavior always works the same way, and if there's a bug, it's going to be basically impossible to spot.

True. I've tested every item the Ai can use at least once, with an understanding of what the behavior of each item was based on the original code and the red rescue team c code. I have checked the 3 places the function is actually called to ensure I understand what the original function was intended to do. That said, it's not enough to ensure there won't be bugs. 
> * Second, AiItemImprovements seems like a very specific patch that I don't see many people wanting to apply. I think it makes more sense as a community skypatch.
Understandable. If others agree, I'll take this out and just include AiItemOptimizations.

Am going to fix the things you suggested anyway, just in case.

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
HappyLappy1: Is 25 in hex (is 0x24 or 36 in the other one), that comment slipped through the cracks, will fix it. This is the number of bytes in a row in the gummi table (which is a half word normally, but a single byte if AddTypes is applied).

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
HappyLappy1: Not sure I understand what you mean by that? I define ``IsAdjacentToEnemy`` in my offsets file(s), so unless that conflicts somehow, this should consistently allocate the same amount of room every time, no?

skytemple_files/_resources/ppmdu_config/pmd2data.xml
Frostbyte0x70: I suppose that could work too. You could also just remove the threshold, what's the problem with the AI using gummies that don't give as much IQ?

HappyLappy1: The vision I had is "the Ai is smart enough to use gummis efficiently on themselves, or has certain flavor preferences"

skytemple_files/patch/handler/ai_item_improvements.py
Frostbyte0x70: Oh, I see. Then this one should not be marked as a `DependantPatch`. And you might also want to state that applying one will overwrite the other or something.

skytemple_files/patch/handler/ai_item_optimizations.py
Frostbyte0x70: Yeah, and also the `depends_on` method.

skytemple_files/_resources/patches/asm_patches/happylappy_asm_mods/ai_item_improvements/common/improvements_ov29.asm
Frostbyte0x70: It's basically just semantics. The only reason why you're using `IsAdjacentToEnemy` in this .area definition is because that function starts right after the area you're using to place your code, right? Not because the `IsAdjacentToEnemy` function has anything to do with that limit. So it makes more sense to say "I can only write until 0x231F358" rather than "I can only write until IsAdjacentToEnemy, because it happens to come right after my area of code".
You could even define a new label like `AreaEnd` with the same offset, to make it clear that it marks the end of your area, regardless of what comes later.

HappyLappy1: I see. I can probably set that up then. There is one instance where I actually am calling that function, but I can change the others to an ``AreaEnd`` or smthn.

skytemple_files/patch/handler/ai_item_improvements.py
```diff
@@ -0,0 +1,200 @@
...
+            elif config.game_region == GAME_REGION_EU:
+                add_types_applied = (
+                    read_u32(rom.loadArm9Overlays([29])[29].data, ADD_TYPES_ADDRESS_EU) != ADD_TYPES_ORIGINAL_CODE
+                )
+            else:
+                raise NotImplementedError()
+        else:
+            raise NotImplementedError()
+        if add_types_applied
+            param["AddTypesApplied"] = 1
```
Chesyon, line 99:
```suggestion
            param["AddTypesApplied"] = 1 # this DOES violate convention, but the ASM code needs to change if AddTypes is applied (for reasons explained above), and this seems to be the only way to pass information from the handler to the ASM code, other than asking the user in the config, which seems redundant when AddTypesApplied can be determined automatically.
```

Hi hello I wrote a lot of the patch handler code, here's the explanation.

HappyLappy1: Tysm chesyon

PR #583, opened 2025-04-26
File/folder deletion support + file/folder root creation support

Adex-8x:
- Adds functionality to delete files and folders from the ROM provided a path
- Adds functionality to create files and folders in the ROM in the filesystem's root folder (e.g., `create_folder_in_rom(rom, 'CUSTOM')` will now work)
- Fixes creating folders in a folderless directory, which would previously assign a folder `firstID` of -2 and render the ROM unable to be saved 
- Fixed a smol typo in an error message

### DIFF
--- a/skytemple_files/common/util.py
+++ b/skytemple_files/common/util.py
@@ -680,7 +680,7 @@ def delete_folder_in_rom(rom: NintendoDSRom, path: str) -> None:
         if current_folder is None:
             raise FileNotFoundError(f(_("Folder {path} does not exist.")))
         for file in current_folder.files:
-            delete_file_in_rom(rom, path + "/" + file)
+            delete_file_in_rom(rom, path + "/" + file, True)
         for folder in current_folder.folders:
             recursive_deletion_folder(path + "/" + folder[0])
         for folder in parent_folder.folders:
--- a/skytemple_files/common/util.py
+++ b/skytemple_files/common/util.py
@@ -705,14 +705,19 @@ def create_folder_in_rom(rom: NintendoDSRom, path: str) -> None:
     found = False
     first_id = -1
     last_child_count = -1
-    for s_name, s_folder in sorted(parent_dir.folders, key=lambda f: f[0]):
-        first_id = s_folder.firstID
-        last_child_count = len(s_folder.files)
-        if s_name > path_list[-1]:
-            found = True
-            break
-    if not found:
-        first_id = first_id + last_child_count
+    folders = parent_dir.folders
+
+    if len(folders) > 0:
+        for s_name, s_folder in sorted(folders, key=lambda f: f[0]):
+            first_id = s_folder.firstID
+            last_child_count = len(s_folder.files)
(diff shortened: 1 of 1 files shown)


skytemple_files/common/util.py
```diff
@@ -603,12 +603,40 @@ def is_binary_in_rom(rom: NintendoDSRom, binary: SectionProtocol | None) -> bool
...
+    folder: Folder | None = rom.filenames.subfolder(dir_name) if len(path_list) > 1 else rom.filenames
+    if target_id is None or folder is None:
+        raise FileNotFoundError(f(_("Folder {dir_name} does not exist.")))
+    folder.files.remove(file_name)
+
+    def recursive_decrement_folder_start_idx(rfolder: Folder, target_idx: int) -> None:
+        if rfolder != folder and rfolder.firstID >= target_idx:
+            rfolder.firstID -= 1
+        for __, sfolder in rfolder.folders:
+            recursive_decrement_folder_start_idx(sfolder, target_idx)
```
Frostbyte0x70, line 625: Can you move this outside this method? You can make it private to signal that it's not used outside this file.

skytemple_files/common/util.py
```diff
@@ -632,16 +660,47 @@ def folder_in_rom_exists(rom: NintendoDSRom, path: str) -> bool:
...
+        if current_folder is None:
+            raise FileNotFoundError(f(_("Folder {path} does not exist.")))
+        for file in current_folder.files:
+            delete_file_in_rom(rom, path + "/" + file, True)
+        for folder in current_folder.folders:
+            recursive_deletion_folder(path + "/" + folder[0])
+        for folder in parent_folder.folders:
+            if folder[0] == current_name:
+                parent_folder.folders.remove(folder)
+                break
```
Frostbyte0x70, line 689: Same as above

Frostbyte0x70 (changes requested):

skytemple_files/common/util.py
Adex-8x: I could move it, but that would entail making an additional parameter for the recursive function to check if `rfolder != folder`. And this is also what an [existing function](<https://github.com/SkyTemple/skytemple-files/blob/feb35608f836bd31c1433e875dce4d3af17785a6/skytemple_files/common/util.py#L620>) does too, so does that need to be changed too, or just this new instance?

skytemple_files/common/util.py
Adex-8x: Also similar reasoning as the earlier comment, but I would also need to pass a NintendoDSRom as a parameter to the recursive function instead of just relying on the scoped variable `rom`, would that be okay?

skytemple_files/common/util.py
Frostbyte0x70: Oh, alright, then we can keep it like that for consistency.
(Also sorry for the delay, I forgor about this lol)

Frostbyte0x70 (approved):
Thanks!

